PR #1, opened 2019-07-05
Create Dockerfile

eddieantonio:
The added Dockerfile creates a small container that will compile and run the ReadAlong-Studio web application. I'm not 100% sure that all functionalities work.


roedoejet:
Thanks @eddieantonio !!

PR #2, opened 2019-07-08
Dev.docker for windows

joanise:
Fixing run.py and Dockerfile to work better in Docker for Windows.


roedoejet:
Great, thanks Eric!

PR #10, opened 2019-09-20
Dev.testing writing tests - don't accept this pull request

joanise:
This pull request is to test continuous integration and code coverage.
I added a test case based on e2e with 4 different xml files and the audio file Aidan recorded.

Please don't accept this pull request, but also don't delete or reject it, just leave it hanging, I want to see what the Coveralls and CI reports look like on it.


roedoejet:
Wow, something very strange is happening with the character escaping. Looks like it is applying it every time it's running. Probably a dictionary mistakenly passed by reference somewhere. I'll look into this on g2p. Looks like the travis & coveralls integration is working alright though...

joanise:
Hi Aidan,

As my first test cases, I created a few tests that simulate what

    readalongs align ap.xml ap.wav out

does, with several variants of ap.xml that I created when I worked on improving the error handling.

I’m created this test PR with my nascent test suite. Don’t accept this yet, I want to work on it further first.

Is there a simpler way to run Coveralls on my code than creating a PR? I’d ideally like to run it myself as I’m writing more tests without having to submit PRs each time.

So far it looks like a nice improvement: +12% for doing 4 end-to-end cases, but I consider this misleading since my test case doesn’t actually verify anything yet, it just runs it. This is partly why I don’t want this PR merged in now!

By the way, is there a way to go further out: I’d like to write a test that actually runs cli.py as if it had been called from the command line, or possibly by actually calling it from the command line, but I don’t see how to integrate that kind of test in the test suite, or especially in the coverage analysis.

Cheers,

Eric

joanise:
The escaping business already exists in the original test suite, it just doesn't get too far in its exponential escaping. On branch `test`, it gets to `\\\\\\\\\\\\\\\?`, which my tests expand to a silly length with 4 or 5 more doublings. Indeed, this needs to be fixed by making sure character escaping only happens once.

roedoejet:
Hey Eric, this all looks great! So, so happy to have you putting your brain on this! :)

Coveralls basically just visualizes the coverage from the report given by [coverage.py](https://coverage.readthedocs.io/en/v4.5.x/). To check the coverage locally I recommend doing the following:

- `pip3 install coverage`
- `cd ~/ReadAlong-Studio/test && coverage run run.py dev`

Typing `coverage run` is basically synonymous with typing `python`, you can give any python file as an argument and it will run it and keep track of the coverage. There are some settings that you need to configure (like which modules it should keep the coverage for). You can add that stuff as flags, but I wrote a config file instead in `~/ReadAlong-Studio/test/.coveragerc`. Feel free to change it as needed, there's good documentation in the link above for coverage.py.

Once you've run the above command, you can view the results on the command line:

`coverage report`

Or you can turn the report into html:

- `coverage html`
- `cd htmlcov && python3 -m http.server`

and it will be on `localhost:8000`

roedoejet:
As for testing the command line, I think it's much easier to do with click than with argparse!

Here's a good link: http://click.palletsprojects.com/en/7.x/testing/

Here are some tests I wrote for the mothertongues CLI that also uses click: https://github.com/roedoejet/mothertongues/blob/master/mtd/tests/test_cli.py

roedoejet:
> The escaping business already exists in the original test suite, it just doesn't get too far in its exponential escaping. On branch `test`, it gets to `\\\\\\\\\\\\\\\?`, which my tests expand to a silly length with 4 or 5 more doublings. Indeed, this needs to be fixed by making sure character escaping only happens once.

Made an issue here: https://github.com/roedoejet/g2p/issues/10

joanise:
I'm done with this pull request, we can close it now.

PR #24, opened 2020-03-17
Dev.temp files on windows using PortableNamedTemporaryFile

joanise:
Tested on Windows only so far. Please test on Mac or Linux. I don't want to merge this if it breaks things on other platforms, and currently I'm unable to run the tests on Linux due to VPN outages.

When running `python run.py dev` in `test/`, I get two failed tests and one error on Windows, which is an improvement over master, where I get 1 failure but three errors, two of which are fixed by this PR:

```
======================================================================
ERROR: test_conversions (test_context_g2p.TestG2P)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\Users\joanise\RAS\ReadAlong-Studio\test\test_context_g2p.py", line 34, in test_conversions
    converter = make_g2p(test['in_lang'], test['out_lang'])
  File "c:\users\joanise\documents\readalong\g2p\g2p\__init__.py", line 27, in make_g2p
    raise(FileNotFoundError("No lang called {in_lang}."))
FileNotFoundError: No lang called {in_lang}.

======================================================================
FAIL: test_reduced_indices (test_context_g2p.TestG2P)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\Users\joanise\RAS\ReadAlong-Studio\test\test_context_g2p.py", line 43, in test_reduced_indices
    (2, 2), (3, 5), (4, 8), (5, 9)])
AssertionError: Lists differ: [(2, 2), (3, 7), (4, 8), (5, 9)] != [(2, 2), (3, 5), (4, 8), (5, 9)]

First differing element 1:
(3, 7)
(3, 5)

- [(2, 2), (3, 7), (4, 8), (5, 9)]
?              ^

+ [(2, 2), (3, 5), (4, 8), (5, 9)]
?              ^


======================================================================
FAIL: testAlignText (test_force_align.TestForceAlignment)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\Users\joanise\RAS\ReadAlong-Studio\test\test_force_align.py", line 48, in testAlignText
    self.assertEqual(xw.attrib['id'], w['id'])
AssertionError: 't0b0d0p0s0w0' != 's0w0'
- t0b0d0p0s0w0
+ s0w0


----------------------------------------------------------------------
Ran 15 tests in 5.228s

FAILED (failures=2, errors=1)
```

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -9,7 +9,7 @@
 #
 #######################################################################
 
-from tempfile import NamedTemporaryFile
+from readalongs.tempfile import PortableNamedTemporaryFile
 from datetime import timedelta
 from typing import List, Union
 import wave
@@ -115,14 +115,14 @@ def align_audio(xml_path: str, wav_path: str, unit:str ='w', bare=False, save_te
     if save_temps:
         dict_file = io.open(save_temps + '.dict', 'wb')
     else:
-        dict_file = NamedTemporaryFile(prefix='readalongs_dict_', delete=False)
+        dict_file = PortableNamedTemporaryFile(prefix='readalongs_dict_', delete=False)
     dict_file.write(dict_data.encode('utf-8'))
     dict_file.flush()
     fsg_data = make_fsg(xml, xml_path, unit=unit)
     if save_temps:
         fsg_file = io.open(save_temps + '.fsg', 'wb')
     else:
-        fsg_file = NamedTemporaryFile(prefix='readalongs_fsg_', delete=False)
+        fsg_file = PortableNamedTemporaryFile(prefix='readalongs_fsg_', delete=False)
     fsg_file.write(fsg_data.encode('utf-8'))
     fsg_file.flush()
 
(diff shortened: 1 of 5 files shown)


roedoejet:
The error and one of the failures are due to g2p, which shouldn't even be tested in this repo anymore, so I'll take those tests out in a separate commit.

I'm not sure what's going on with the other error, other than it looks like it's not creating the correct XML. I looked at the Travis log though, and these tests have actually been failing for a while. But, annoyingly in our test runner it was never set to exit with an error if the tests fail. I'll also add that in a separate commit.

I ran on my Mac and it runs (but with the errors/failures you mention). Looking at the Travis log (linux) it's also replicating the same behaviour, so I'll merge this.

roedoejet:
I made the changes mentioned above here: bf6b1f3db27ce4ab7cad2673c1cf7e779a561f51

Notice that the build is now failing as of e681fb84aa5e68ca310f2f9adc1c8e3dff0f9bb3 but this is actually what we want so long as that test is still failing.

joanise:
Thanks for the testing! The testAlignText failure has been bugging me for a while, I should have reported it a long time ago.
And I guess it makes sense to disable the g2p tests if those are no longer testing the current code.

PR #33, opened 2020-07-08
do-not-align audio

roedoejet:
As discussed, here's the PR for the do-not-align audio section, along with a basic implementation of the configuration


readalongs/align.py
```diff
@@ -33,46 +33,58 @@
...
+    """ Given the start and end of a segment and a list of do-not-align segments,
+        If one of the do-not-align segments occurs inside one of the start-end range,
+        align the start or end with the do-not-align segment, whichever requires minimal change
+    """
+    if isinstance(start, float):
+        LOGGER.info(f'Converting {start} to milliseconds')
+        start = int(start * 1000)
+    if isinstance(end, float):
+        LOGGER.info(f'Converting {end} to milliseconds')
+        end = int(end * 1000)
```
joanise, line 47: This seems dangerous to me. It means that 100 is time=.1s, but 100.0 is time 100s. That's going to be a serious gotcha! If we want to support both s and ms, I would rather have different variable names for them...

readalongs/align.py
```diff
@@ -33,46 +33,58 @@
...
-    
+
     Parameters
     ----------
     xml_path : str
         Path to XML input file in TEI-like format
-    wav_path : str
-        Path to audio input (WAV or MP3)
+    audio_path : str
+        Path to audio input
```
joanise, line 80: It might still be useful to put an indication here as to which audio file formats are supported. Maybe "any format supported by X" where X is the library we're using?

readalongs/align.py
```diff
@@ -179,6 +217,13 @@ def frames_to_time(frames):
     for seg in ps.seg():
         start = frames_to_time(seg.start_frame)
         end = frames_to_time(seg.end_frame + 1)
+        if do_not_align_segments and method == 'remove':
+            start += (calculate_adjustment(start,
+                                           do_not_align_segments) / 1000)
+            end += (calculate_adjustment(end, do_not_align_segments) / 1000)
+            start, end = correct_adjustments(start, end, do_not_align_segments)
+            start /= 1000
+            end /= 1000
```
joanise, line 226: I don't like this mucking around with seconds and ms. Can we pick a consistent unit to use everywhere, and only convert (if necessary) when data is read from or written to file? Ideally, I would prefer having this function here, as well as calculate_adjustment and correct_adjustment all work in the same unit, probably the ms.

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
+@click.option("-c", "--config", type=click.Path(exists=True), help="Use ReadAlong-Studio configuration file")
+@click.option("-C", "--closed-captioning", is_flag=True, help="Export sentences to WebVTT and SRT files")
+@click.option("-d", "--debug", is_flag=True, help="Add debugging messages to logger")
+@click.option("-f", "--force-overwrite", is_flag=True, help="Force overwrite output files")
+@click.option("-i", "--text-input", is_flag=True, help="Input is plain text (assume paragraphs separated by blank lines, 1 paragraph per page)")
+@click.option("-l", "--language", type=click.Choice(LANGS, case_sensitive=False), help="Set language for plain text input")
+@click.option("-u", "--unit", type=click.Choice(["w", "m"], case_sensitive=False), help="Unit (w = word, m = morpheme) to align to")
+@click.option("-s", "--save-temps", is_flag=True, help="Save intermediate stages of processing and temporary files (dictionary, FSG, tokenization etc)")
+@click.option("-t", "--text-grid", is_flag=True, help="Export to Praat TextGrid & ELAN eaf file")
+@click.option("-x", "--output-xhtml", is_flag=True, help="Output simple XHTML instead of XML")
```
joanise, line 90: I see, here and elsewhere, that you standardized strings to double quotes. I searched and found online some advice saying this is fairly common convention, but do we have a style guide we're trying to follow?
Since Python doesn't care about single or double quotes, I have not been consistent about them myself, but I would be happy to have a follow a style guide on this.

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
 def align(**kwargs):
-    """Align INPUTFILE and WAVFILE and create output files at OUTPUT_BASE.
+    """Align INPUTFILE and AUDIOFILE and create output files at OUTPUT_BASE.
 
     inputfile : A path to the input text file
 
-    wavfile : A path to the input audio file
+    audiofile : A path to the input audio file
```
joanise, line 96: Here's it's more important to tell the user what audio files are supported, if we can, since this string is part of the help message.

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
 
     output-base : A base name for output files
     """
-    if os.path.exists(kwargs['output_base']):
-        if not kwargs['force_overwrite']:
+    config = kwargs.get("config", None)
+    if config and config.endswith("json"):
```
joanise, line 101: Hum, this really needs an else clause: if config exists and is not `.json`, we have to issue a `click.UsageError()`!

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
-@click.option('-s', '--save-temps', is_flag=True,
-              help='Save intermediate stages of processing and temporary files (dictionary, FSG, tokenization etc)')
-@click.option('-t', '--text-grid', is_flag=True, help='Export to Praat TextGrid & ELAN eaf file')
-@click.option('-x', '--output-xhtml', is_flag=True, help='Output simple XHTML instead of XML')
+@app.cli.command(context_settings=CONTEXT_SETTINGS, short_help="Force align a text and a sound file.")
+@click.argument("inputfile", type=click.Path(exists=True, readable=True))
+@click.argument("audiofile", type=click.Path(exists=True, readable=True))
+@click.argument("output-base", type=click.Path())
+@click.option("-b", "--bare", is_flag=True, help="Bare alignments do not split silences between words")
+@click.option("-c", "--config", type=click.Path(exists=True), help="Use ReadAlong-Studio configuration file")
```
joanise, line 81: Add `(in JSON format)` at the end of the `help` string.
```suggestion
@click.option("-c", "--config", type=click.Path(exists=True), help="Use ReadAlong-Studio configuration file (in JSON format)")
```

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
     """
-    if os.path.exists(kwargs['output_base']):
-        if not kwargs['force_overwrite']:
+    config = kwargs.get("config", None)
+    if config and config.endswith("json"):
+        try:
+            with open(config) as f:
+                config = json.load(f)
+        except json.decoder.JSONDecodeError:
+            LOGGER.error(f"Config file at {config} is not valid json.")
```
joanise, line 106:
```suggestion
    if config:
        if config.endswith("json"):
            try:
                with open(config) as f:
                    config = json.load(f)
            except json.decoder.JSONDecodeError:
                LOGGER.error(f"Config file at {config} is not valid json.")
        else:
            raise click.BadParameter(f"Config file '{config}' must be in JSON format")
```

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
-        tokenized_xml_path = '%s.xhtml' % output_base
+        os.mkdir(kwargs["output_base"])
+    output_base = os.path.join(
+        kwargs["output_base"], os.path.basename(kwargs["output_base"])
+    )
+    if kwargs["debug"]:
+        LOGGER.setLevel("DEBUG")
+    if kwargs["text_input"]:
+        if not kwargs["language"]:
+            LOGGER.warn(
```
joanise, line 122: I'm really not that keen on adding these extra newlines, like here, and above where you have `(` on one line, the `f"..."` string on another, and `)` on yet another. I'm not going to fight it, but I hope you don't mind if I don't systematically reproduce it in my own code...

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
+    if os.path.exists(smil_path) and not kwargs["force_overwrite"]:
+        raise click.BadParameter(
+            "Output file %s exists already, did you mean to do that?" % smil_path
+        )
+    _, audio_ext = os.path.splitext(kwargs["audiofile"])
+    audio_path = output_base + audio_ext
+    if os.path.exists(audio_path) and not kwargs["force_overwrite"]:
+        raise click.BadParameter(
+            "Output file %s exists already, did you mean to do that?" % audio_path
+        )
```
joanise, line 149:
```suggestion
        raise click.BadParameter(
            "Output file %s exists already, use -f to overwrite."
            % tokenized_xml_path
        )
    smil_path = output_base + ".smil"
    if os.path.exists(smil_path) and not kwargs["force_overwrite"]:
        raise click.BadParameter(
            "Output file %s exists already, use -f to overwrite." % smil_path
        )
    _, audio_ext = os.path.splitext(kwargs["audiofile"])
    audio_path = output_base + audio_ext
    if os.path.exists(audio_path) and not kwargs["force_overwrite"]:
        raise click.BadParameter(
            "Output file %s exists already, use -f to overwrite." % audio_path
        )
```

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
     unit = kwargs.get("unit", "w")
     bare = kwargs.get("bare", False)
-    if not unit:     # .get() above should handle this but apparently the way kwargs is implemented
-        unit = "w"   # unit could still be None here.
+    if (
+        not unit
+    ):  # .get() above should handle this but apparently the way kwargs is implemented
```
joanise, line 154: OK, I really don't like this! Why break such a short statement over three lines?

readalongs/cli.py
```diff
@@ -66,116 +73,132 @@ def cli():
...
-                              save_temps=(output_base
-                                          if kwargs['save_temps'] else None))
+        results = align_audio(
+            kwargs["inputfile"],
+            kwargs["audiofile"],
+            unit=unit,
+            bare=bare,
+            config=config,
+            save_temps=(output_base if kwargs["save_temps"] else None),
+        )
```
joanise, line 164: This does improve legibility, so I agree with the extra lines here, when there is a good legibility benefit.

test/test_audio.py
```diff
@@ -0,0 +1,155 @@
+import os
```
joanise, line 1: Nice TDD work in this file! I like it.

readalongs/align.py
roedoejet: Yeah, I'm not a big fan of this back and forth either. The problem is that the library we're using to manipulate audio (pydub) works in milliseconds and the SMIL files have all used seconds. I'm not sure if the SMIL can be changed, but I don't really anticipate this being a gotcha - so long as we know that internally we use milliseconds and in the SMIL we print to seconds we should be fine - there's no scenario where someone should be manually entering anything here. All that said, I would certainly be in favour of just doing everything in milliseconds if that would be possible.

readalongs/align.py
roedoejet: Good idea! We're using pydub which I believe just uses the system installation of ffmpeg, so we can put that there.

readalongs/cli.py
roedoejet: It doesn't matter, and it's not enforced by pep8. I use double quotes just to be consistent, but I think ideally what we should do is commit to an auto-formatting tool like (black)[https://black.readthedocs.io/en/stable/the_black_code_style.html] and all agree on using that. Or adding a commit hook to run the formatting on before pushing or something.

readalongs/cli.py
roedoejet: I think this goes back to deciding on an auto-formatting tool. This was done by `black` on my machine

readalongs/cli.py
roedoejet: I'm not sure how this happened!

test/test_audio.py
roedoejet: Thanks! Yes, it was nice to have a bite-sized chunk to work on properly

readalongs/align.py
roedoejet: OK, I refactored it (18b713d) so that `calculate_adjustment` and `correct_adjustments` are both getting passed timestamps in ms. Let me know if you think my change is an improvement.

readalongs/align.py
roedoejet: see 18b713d

joanise: Yeah, that's much better. Calling it `*_ms` makes it nicely self-documenting, and adding the comments in the functions.
We can't remove having to deal with both, but this makes it clear.

readalongs/cli.py
joanise: I bet Black split it because the original line was longer that 88 characters when you consider the length of the comment.
I'm willing to accept such weird things if we're agreeing to an automated formatting tool.

readalongs/align.py
```diff
@@ -33,46 +33,58 @@
...
-# reading methods being context managers; the following checks whether the
-# necessary methods are present and if not, add thems.
-#
-# Based on http://web.mit.edu/jgross/Public/21M.065/sound.py 9-24-2017
-####
-
 
-def _trivial__enter__(self):
-    return self
+def correct_adjustments(start: Union[int, float], end: Union[int, float], do_not_align_segments: List[object]) -> (int, int):
```
joanise, line 37: Now that this function is just in ms, do we still need `Union[int, float]`? The Black doc suggests that this is equivalent to just `float`, so I would switch it, as long as the code still works when we do.

PR #35, opened 2020-07-15
Dev.do not align text

joanise:
do-not-align text feature is implemented and ready to be pulled in.

End-to-end testing is not extensive yet, but I want to merge this branch in before continuing.


joanise:
Hum, I probably could have pushed the first two commits to master directly, they're not relevant to this PR, but never mind, they can ride along, it's not harmful.

readalongs/align.py
```diff
@@ -462,19 +462,21 @@ def convert_to_xhtml(tokenized_xml, title='Book'):
 
 TEI_TEMPLATE = """<?xml version='1.0' encoding='utf-8'?>
 <TEI>
+    <!-- To exclude a page (div), paragraph (p), or sentence (s), from alignment, add
```
roedoejet, line 465: Nice, this is a good idea to document inside the template.

readalongs/cli.py
```diff
@@ -219,6 +232,9 @@ def epub(**kwargs):
               required=True, help='Set language for input file')
 def prepare(**kwargs):
     """Prepare XMLFILE for 'readalongs align' from plain text INPUTFILE.
+    INPUTFILE must be plain text encoded in utf-8, with one sentence per line,
```
roedoejet, line 235: Hm, `create_input_tei` doesn't tokenize words, but what if you want to just remove a single word from the aligner? Can you wrap a word in a `<w do-not-align=true>foobar</w>` or something? This has come up before where it's individual words that need to be omitted. Not sure if we would ever need subword, but if we did, is there a way to handle that too?

test/test_dna_text.py
```diff
@@ -0,0 +1,142 @@
...
+<s id="s1"><w id="s1w0">Voici</w> <w id="s1w1">une</w> <w id="s1w2">deuxième</w> <w id="s1w3">phrase</w>.</s>
+</document>"""
+        self.assertEqual(ids_as_txt, ref_with_ids)
+
+    def test_tok_some_words(self):
+        txt = """<document xml:lang="fra">
+<p><s>Bonjour! Comment ça va?</s></p>
+<p do-not-align="true"><s>Bonjour! Comment ça va?</s></p>
+<s do-not-align="TRUE">Voici une deuxième phrase.</s>
+<s>Un <foo do-not-align="1">mot ou deux</foo> à exclure.</s>
```
roedoejet, line 54: nvm, the question of omitting words is answered here. Maybe the template should be updated then as it only mentions `<s>` and `<p>` and `<div>` elements?

readalongs/cli.py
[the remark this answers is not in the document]
joanise: I guess that's a good idea. My philosophy is that I don't really want to force the user to use a specific extension, letting them use whatever they prefer, but it's only a mild preference, I can change it.

joanise: Subword: I don't think we can handle that.
Word within one sentence: I believe the current behaviour is that `<s>blah <foo do-not-align="true">bar</foo> blah</s>` will result in blah blah being passed to the forced aligner, and bar not. It seems fine to me that bar does not get tokenized, since it's being excluded from alignment. But I'll have to do more tests to see if it makes sense the way I have it.

If you have `<w do-not-align="true">bar</w>`, that'll cause problems, because later one the tokenizer will see there is already a `w` element a decide it doesn't need to tokenize anything. So that cannot be allowed, which is something else I should probably handle: look for a report this as an error.

test/test_dna_text.py
joanise: Yeah, that's a good solution.

readalongs/cli.py
littell: The tokenize_xml function should manage to tokenize things into <w>'s even when wrapped in an arbitrary element; I just checked and it appears to work.  Have we tried it end-to-end and looked at the intermediate representation, to see if <foo do-not-align="true"><w>bar</w></foo> is being created or not?

joanise: @littell The solution I chose to implement DNA is that I skip tokenization whenever an element is marked as DNA. This seemed like the right thing to do to me, so I'm explicitly trying to prevent that `bar` from being wrapped in a `<w>`.
Part of my logic is that DNA text might be in a different language, or whatever else, and it might not be helpful or appropriate to tokenize.
I also prevent the DNA items from getting ids, and in some ways I'm cheating: I relying on the fact that later processing only extracts the sequence to words with ids in them.
I'm not sure what would happen if I had `<w>` element without an id, but I need to do something cleaner with that because right now a pre-existing `<w>` element in the input causes trouble.

joanise: @littell  Just did the end-to-end test, and I will add it to unit testing: if the input has
`<s>some text <foo do-not-align="true">exclude</foo> more text</s>`
then the tokenized.xml file has `<w>` elements around some, text, more, and text, but not around exclude.

If the input has 
`<s>some text <w do-not-align="true">exclude</w> more text</s>`
then you get these two messages and a trace dump:
```
INFO - Words (<w>) already present; skipping tokenization
ERROR - w-type element without id in file data/ej-fra-dna.xml
Traceback (most recent call last):
...

I think the idea behind `INFO - Words (<w>) already present; skipping tokenization` is that if an xml input file already has even one `<w>` element in it, we assume it's already been tokenized and should not be retokenized. DNA interacts poorly with that because I prevent assignment of an id to that word, in this case, so I should issue a user error message when the is a `<w>` inside a DNA element at any level.

```

joanise: What I have not yet tested is how the web component handles
`<s><w id=...>some</w> <w id=...>text</w> <foo do-not-align="true">exclude</foo> <w id=...>more</w> <w id=...>text</w></s>`
and unfortunately I am currently working on the GPSC where I cannot actually spin up a web server, so if someone else could test that it would be helpful.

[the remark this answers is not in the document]
joanise: Done, I now add .xml if the file does not end in .xml.

joanise: Fixed: a <w> element inside a DNA section now raises a RuntimeError, which triggers a user error message.

joanise:
I think I've addressed everything I need to address now, see my many comments above and a few commits. I'd like to merge now and move on to the Black reformatting business.

joanise:
@roedoejet  My last commit above adds 15 seconds to the unit testing, and therefore to every build. Are you OK with that?

PR #38, opened 2020-08-28
Dev.contributing.md

joanise:
Please review my new `Contributing.md` documentation. Is it the right level for non-core members? Anything important missing or unclear?


dhdaines:
This looks good although I haven't tried everything out that is mentioned in it.  One basic thing that maybe should be mentioned is the minimum version of Python, I guess we are probably requiring at least 3.7, right?

joanise:
@dhdaines We require Python 3.6 in setup.py. I forget why, but at some point we decided we didn't want to exclude 3.6, possibly for CI? Aidan might remember.
For testing all this, could you enable both types of Git hooks and see if they work for you with my instructions?

joanise:
@littell Your review would be appreciated because I understand you haven't used such hooks before (am I right?). So you might be able to catch if it answers all the questions a new contributor might have.

Contributing.md
```diff
@@ -0,0 +1,137 @@
...
+hooks above, but please take a moment to do this:
+
+- If you don't already use Node or nvm, install nvm in your ~/.nvm folder:
+```sh
+wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash`
+```
+This will add a few lines to your `.bashrc` file, which you'll need to execute now,
+possibly by starting a new shell.
+
+- Install Node, preferably version 12.18.3
```
roedoejet, line 116: I feel like the docs here can just use `nvm install node`. I only said to use 12.18.3 because it's the current version I'm working on. But, since we use basically no JavaScript except for the commitlint package, I think we're fairly safe just recommending the latest stable version. Plus it means they don't have to do `nvm use 12.18.3`, so it's one less step

Contributing.md
```diff
@@ -0,0 +1,137 @@
...
+- In your ReadAlong/Studio sandbox, install the husky commit-msg hook using npm, the node
+  package manager you just installed using nvm. The file `package.json` in Studio is what
+  tells npm to install husky as a pre-commit hook, and also what tells husky to invoke
+  commitlint on your commit messages.
+```sh
+npm install
+```
+
+- Now, next time you make a change and commit it, your commit log will be checked:
+  - `git commit -m'non-compliant commit log text'` outputs an error
```
roedoejet, line 136: hehe, I would write this `git commit -m 'non-compliant commit log text'` but I noticed you doing it without the space on the command line so I think you meant to write it this way. I don't care, but maybe this is another tabs vs spaces type argument.

roedoejet: i.e. we should add a pre-commit hook that fails if you write it like this `git commit -m'non-compliant commit log text'` but also fails if you write it like this `git commit -m 'non-compliant commit log text'` 😛

README.md
```diff
@@ -58,8 +58,8 @@ The best thing to do is install with pip `pip install readalongs`.
 Otherwise, clone the repo and pip install it locally.
 
 ```sh
-$ git clone https://github.com/dhdaines/ReadAlong-Studio.git
-$ cd ReadAlong-Studio
+$ git clone https://github.com/Readalongs/Studio.git
```
roedoejet, line 61:
```suggestion
$ git clone https://github.com/ReadAlongs/Studio.git
```

littell:
The 3.6 is because the last Ubuntu LTS assumes 3.6, and there was no
official 3.7 distro for Ubuntu, so users would have to install a whole new
Python. And googling about it could get you some bad advice that can break
your Ubuntu distro in ways that you don't notice until later. So a
potential quagmire, and I don't think there was any 3.7 functionality we
actually needed.

(There may now be an official 3.7, but the more convoluted route is still
top results when you search for it.)

Contributing.md
dhdaines: what is nvm? do you mean npm?

joanise:
@littell Thanks for those 3.6 details, I'm glad you remember! So let's keep it to 3.6, for sure, and I'm glad it's documented in this PR, too.

Contributing.md
```diff
@@ -0,0 +1,137 @@
...
+    optional footer
+
+Valid types: (these are the default, which we're using as is for now)
+ - build: commits for the build system
+ - chore: maintain the repo, not the code itself
+ - ci: commits for the continuous integration system
+ - docs: adding and changing documentation
+ - feat: adding a new feature
+ - fix: fixing something
+ - improvement: improve something
```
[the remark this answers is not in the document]
joanise: I don't have a specific use case, so I don't mind taking it out, but I think we have it right now, because `commit -m'foo: bar'` gives me this error message:
```
✖   type must be one of [build, chore, ci, docs, feat, fix, improvement, perf, refactor, revert, style, test] [type-enum]
```
with `improvement` included in the list.

Contributing.md
joanise: @dhdaines I was confused by nvm and npm, and got it wrong half the time when Aidan told me what to do...
nvm is the virtual environment manager for node/js that lets you have multiple versions active at the same time. `nvm use <version>` lets you activate the one you want.
npm is the package manager to install stuff, once you've picked the V env you want.

joanise: @roedoejet For sure, if the latest node will do, that's what we should document and use. I'll test it that way and update my doc if it works.

Contributing.md
joanise: Indeed, I never put a space there! I don't care either way how we show it here, although the second example, which I say works, also doesn't have that space.
Do you expect it might confuse some readers?

Contributing.md
joanise: OK, tested, I am taking out the specific version and the `nvm use` statement.

Contributing.md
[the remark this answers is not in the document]
joanise: Hum, I see that I have a `commitlint.config.js` file in my sandbox, but I don't remember where it came from. It doesn't include "improvement", but for some reason this config file is ignored and doesn't get taken into account.

joanise:
@roedoejet Is this good enough to merge in, or is there more to do? Ideally, I'd like to figure out why my `commitlint.config.js` gets ignored while it seems yours does not...

Contributing.md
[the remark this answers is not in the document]
roedoejet: yeah, that's strange, where is it getting that from?? in package.json it clearly shows it's extending https://www.npmjs.com/package/@commitlint/config-conventional

roedoejet:
Go for it, @joanise

PR #41, opened 2020-11-27
Fix LANGS list in ReadAlongs align and views

joanise:
Fixes https://github.com/roedoejet/g2p/issues/70

`LANGS_AVAILABLE` defined in `g2p` does not list all languages, so don't use it to define `LANGS` in ReadAlongs/Studio.

Tested for readalongs align (changes in `cli.py`), but not for `views.py`, where the same list is used, and where I therefore believe the change was also needed. The unit test suite does not exercise the code in `views.py` that uses `LANGS`, namely `step()`, and I don't know how to exercise it, so that should ideally get tested before merging this PR in.

This patch is required to successfully run `readalongs align` on any Michif data, since `crg-dv` and `crg-tmd` were not found in LANGS before this patch.

### DIFF
--- a/readalongs/cli.py
+++ b/readalongs/cli.py
@@ -27,8 +27,6 @@ from tempfile import TemporaryFile
 
 import click
 from flask.cli import FlaskGroup
-from g2p.mappings.langs import LANGS_AVAILABLE, LANGS_NETWORK
-from networkx import has_path
 
 from readalongs._version import __version__
 from readalongs.align import (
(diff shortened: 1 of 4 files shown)


joanise:
Currently, my code supports this list of languages:

`-l, --language [alq|atj|ckt|crg-dv|crg-tmd|crj|crj-norm|crk|crk-no-symbols|crl|crl-norm|crm|crm-norm|csw|csw-norm|ctp|dan|eng|fra|git|git|git|gla|iku|kkz|kwk-boas|kwk-napa|kwk-umista|kwk-umista|lml|moh|moh|oji|oji-syl|see|srs|str|tce|tce-norm|tgx|tli|tli-norm|und|win]`

Question: do we want to filter out `*-norm`? I think yes, we do, and maybe also `crk-no-symbols`.

dhdaines:
definitely would want to filter out -norm! not sure what crk-no-symbols is but guessing it is also an intermediate representation (did we ever figure out if we could mark these as such, and filter them out of the list that way?)

joanise:
`crk-no-symbols` is `crk` wtih "&" replaced by how it's pronounced, namely "êkwa". So yes, I want to filter it out too.

I'll push a commit filtering *-norm and crk-no-symbols.

joanise:
@dhdaines  I agree it would be better to flag nodes in the language network as internal intermediates, but that's a bigger project I'm not up to undertaking just now. I just pushed a simpler solution that filters them out in Studio instead.

roedoejet:
So I think the studio web app is working with your changes now as of 
bfdb5cc. I'm fine to merge this.

joanise:
@roedoejet Yeah, I noticed I broke views.py because I create a list of lang codes, but it expected a list of of code:name pairs.

If you don't mind, I'd like to clean up your quick patch a bit, I think we can have the code much more concise. I don't suppose there's an easy to way to unit-test the studio app easily, is there...

joanise:
Actually, never mind, on further analysis, @roedoejet , I like your code as is. The change I had in mind wouldn't be better.

PR #45, opened 2021-01-22
Dev.images

roedoejet:
This addresses the commit in the "Support adding images by page in config.json" task on clickup

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -25,6 +25,7 @@ from webvtt import Caption, WebVTT
 from readalongs.audio_utils import mute_section, read_audio_from_file, remove_section
 from readalongs.log import LOGGER
 from readalongs.tempfile import PortableNamedTemporaryFile
+from readalongs.text.add_elements_to_xml import add_images, add_supplementary_xml
 from readalongs.text.add_ids_to_xml import add_ids
 from readalongs.text.convert_xml import convert_xml
 from readalongs.text.lang_id import add_lang_ids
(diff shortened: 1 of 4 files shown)


joanise (changes requested):
This work looks good. Very quickly done, and handles the two cases we suggested. Excellent!

However, I would like to see some more unit testing for problem situations. I think this code will produce unintelligible exceptions if the user input in the config file is not valid, e.g., if the images is not an array of pairs, or who knows how else someone might do it wrong. Similar issue if the provided XML string is wrong. Also, this unit test does not test from an actual config.yaml file, only from the internal function calls. 

We certainly don't have to extensively test all possible errors and situations, but I would suggest at least:
 - one read from config.yaml test,
 - one invalid "images" test, and
 - one invalid XML test.

Thanks for writing this feature so fast!

roedoejet:
Hey @joanise - thanks for the review! These days I prefer to do things right away if I have a few moments! I agree with your suggestion for the invalid images and xml tests, I've added some for those (although an xpath that doesn't return any elements should just be a warning I reckon, which is what I implemented). I'm not sure I agree with the config test (also, do you mean config.json or what is this config.yml?). It's a bulky test that requires additional files. It's arguably better included as part of our integration testing, otherwise I'm just unittesting the builtin `json` module's ability to read in json. I suppose we could have a unit test that just reads in a config file and raises an error if it's not valid json, but testing the interactions between reading the file and the current unittests doesn't seem necessary. Thanks again for the review, your suggestions make this PR more robust!

joanise:
Yes, I mean config.json, you're right. It's OK to leave it out, but we should make sure to have integration testing somewhere, which we don't yet. We can merge this PR as is, thanks for adding the error checking and user error messages, it'll be a lot friendlier to use.

roedoejet:
agreed! We should definitely have some dedicated integration testing between all of these different modules. Thanks for the review and merge @joanise !

PR #50, opened 2021-03-12
Refactor tokenizer to g2p

joanise:
Now that the tokenizer has been migrated to g2p, remove it from Studio and use the version in g2p.


joanise:
Global code coverage goes down because some previously-covered code is now in g2p and no longer counts in the coverage analysis. Nonetheless, working to improve coverage, in particular by removing redundant code.

PR #54, opened 2021-05-04
deps: bump networkx to 2.5

eddieantonio:
The version specified in requirements.txt used to import `gcd` from `fractions` (it's actually in `math`), and thus, crashed on startup. networkx 2.5 includes this commit which fixes the problem: https://github.com/networkx/networkx/commit/b007158f3bfbcf77c52e4b39a81061914788ddf9#diff-21e03bb1d46583650bcad6e960f2ab8a5397395c986942b59314033e963dd3fc

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ flask-socketio==4.3.2
 g2p>=0.5.*
 iso-639==0.4.5
 lxml==4.6.3
-networkx==2.3
+networkx==2.5
 numpy>=1.16.4
 panphon>=0.14
 soundswallower==0.1.1


joanise:
Works fine with 2.3 or 2.5 for me, but if 2.3 is causing problems for you, I don't mind bumping it up.

eddieantonio:
> Works fine with 2.3 or 2.5 for me, but if 2.3 is causing problems for you, I don't mind bumping it up.

I think it breaks on 2.3 for a version of Python that doesn't _happen_ to do `from math import gcd` in `fractions.py`. Python 3.9 does not do this, hence it crashes on this version of Python on startup.

PR #55, opened 2021-05-04
fix: normalize key to NFKC in lexicon

eddieantonio:
This might be a cursed commit.

For some reason that is not obvious to me, the CMU lexicon is normalized in NFKC form (why?????). That means that text that is looked up must _also_ be in NFKC form. I believe most other text flowing through the system is in NFD form? Either way, if the cmudict file _happens_ to gain non-ASCII keys — I don't know who would do such a thing 😉 — this commit will prevent it from crashing!

### DIFF
--- a/readalongs/text/lexicon_g2p.py
+++ b/readalongs/text/lexicon_g2p.py
@@ -91,6 +91,7 @@ class LexiconG2P:
         len_text = len(text)
         text = text.strip("#").strip()
         text = text.upper()
+        text = normalize("NFKC", text)
         if text not in self.entries:
             raise KeyError(
                 "Word %s not found in lexicon %s" % (text, self.metadata["src"])


joanise:
Thanks for that PR, @eddieantonio. I'll have to think about how we want to make sure no future problems occur. I want to catch the case where the dict file is edited in the future and something is inserted in a different normalization: either some unit test should fail, and thus warn us, or the file should be converted as it's loaded.

This file is quite an old one that's been carried over from an old CMU project, so that's probably why it uses NFKC forms.

joanise:
Oh, never mind, I just saw that when we load the cmudict file, we also apply NFKC normalization, so your patch should be applicable without any risk. The file itself is in pure 7-bit ASCII, without even any lower case letters, but I agree with making the code robust to other input.

joanise:
As for non-ASCII keys, well, English does borrow some foreign words with their accents, sometimes. "À la carte" comes to mind.

PR #59, opened 2021-05-28
Dev.g2p cascade

joanise:
g2p now optionally cascades through several languages if there are errors with the language specified in the xml file.

Also, und now gets pre-processed with unidecode, so it handles all letters in all languages, based on how they're described in the Unicode standard.


littell (approved):
Looks good to me

joanise:
@littell Thanks for the review!

PR #64, opened 2021-07-03
feat: Misc utility scripts that are peripheral to ReadAlongs Studio

joanise:
For a while, I've been trying to figure out how to push `syll_parse.py` and `non-caching-server.py` to GitHub in a way that respects there respective original licenses yet makes them available for optional use with ReadAlongs/Studio. This is my proposed solution: a directory called `misc-utils` where each file has its own individual licensing information right in the source code, and a description of the NRC modifications. Plus a `README.md` that says what these are and what they are used for.

I want review from Eddie Antonio Santos, Fineen Davis, Aidan Pine and Pat Littell for this PR, to make sure we're all OK with the approach I'm taking here.

Deviation for our normal model is necessary because these files are derivatives of other OSS work.

### DIFF
--- a/misc-utils/README.md
+++ b/misc-utils/README.md
@@ -1,0 +1,31 @@
+## Misc Utils
+
+This directory contains miscellaneous utility scripts that are not part of the
+core of ReadAlongs Studio, but help with specific tasks in the perifery.
+
+Their Copyrights are different from the rest of the repo, and declared with
+license in each file, along with a documentation of the original provenance of
+the files.
+
+### `syll_parse.py`
+
+Syllabify a tokenized readalongs XML file, so that highlighting can happen
+syllable by syllable instead of word by word. This script should work reasonably
+well for languages in the latin script where syllabification follows the
+Sonority Sequencing Principle (SSP). For example, this works on Algonquin.
+
+Must be called manually after
+readalongs tokenize and before readalongs align or readalong g2p:
+
+    readalongs prepare -l my_lang file.txt file.xml
+    readalongs tokenize file.xml file-tok.xml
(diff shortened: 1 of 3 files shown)


eddieantonio (approved):
This looks good to me, HOWEVER:

As far as I understand, browsers that have some old bit of Webkit code (so Safari, Chrome, and Edge may all have this code) make assumptions about the caching policies based on incomplete/ambiguous caching headers sent by the a server, e.g., Python's simple HTTP server. Basically, the built-in `http.server` tells the browser ["hey, this resource was last modified at this time"](https://github.com/python/cpython/blob/1b133ab8412e4f6a53914212df7a7656f376256d/Lib/http/server.py#L750-L751), and given no other information, [the browsers choose to store "static" assets in a disk cache and  does not attempt to request it again for a algorithmically determined time](https://webmasters.stackexchange.com/a/111299).

Frankly, this issue about caching can be solved by using a “real” server, instead of this _or_ `python3 -m http.server`.

If you want to avoid maintaining this file, perhaps we can find some alternative lightweight server that sends correct caching headers? I use `sirv-cli` that is installable via npm:

    npm install -g sirv-cli

And then invoke with the `--dev` command:

    sirv --dev

...which has [caching workarounds](https://github.com/lukeed/sirv/blob/0c832aac6e3eca4c242d0f7d07de2da1abd3d547/packages/sirv/index.js#L47) like this code.

eddieantonio (approved):

joanise:
@eddieantonio Thanks for your review! You're right, this server is minimal and not great, but the point of using it is to be able to run some quick tests without having to install anything. While there are many other better solutions, this is the one I like to use because I don't have to fight with anything, it just works (mostly).

PR #67, opened 2021-07-26
doc: add note about Python version

eddieantonio:
Document the minimum version of Python required. Thanks @tobyatgithub!

### DIFF
--- a/readalongs/python_version.py
+++ b/readalongs/python_version.py
@@ -1,0 +1,27 @@
+"""
+Specifies the minimum version of Python required for this project.
+"""
+
+import sys
+
+# Please write this as a tuple of (major, minor, patch) verison:
+MINIMUM_PYTHON_VERSION_REQUIRED = (3, 7, 0)
+
+
+def ensure_using_supported_python_version():
+    if sys.version_info < MINIMUM_PYTHON_VERSION_REQUIRED:
+        raise Exception(
+            "Python "
+            + _version_to_str(MINIMUM_PYTHON_VERSION_REQUIRED)
+            + " required (you are currently using Python "
+            + _version_to_str(sys.version_info)
+            + "). Please use a newer version of Python."
+        )
+
+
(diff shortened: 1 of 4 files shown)


joanise (changes requested):
@eddieantonio Thanks for doing this, it should definitely be documented! I'm sure the answer to my next question is yes: is there a line of code we could add to `cli.py` and `app.py` to say "requires 3.7" that would exit immediately with a clear error message in older Pythons? (My syntax is approximate Perl, I'm sure the same exists in Python.)

eddieantonio:
> @eddieantonio Thanks for doing this, it should definitely be documented! I'm sure the answer to my next question is yes: is there a line of code we could add to `cli.py` and `app.py` to say "requires 3.7" that would exit immediately with a clear error message in older Pythons? (My syntax is approximate Perl, I'm sure the same exists in Python.)

It is indeed possible, though doesn't have neat, straightforward syntax like Perl (yes, I just complimented Perl's syntax!)

readalongs/python_version.py
```diff
@@ -0,0 +1,27 @@
...
+
+def ensure_using_supported_python_version():
+    if sys.version_info < MINIMUM_PYTHON_VERSION_REQUIRED:
+        sys.exit(
+            "Python "
+            + _version_to_str(MINIMUM_PYTHON_VERSION_REQUIRED)
+            + " required (you are currently using Python "
+            + _version_to_str(sys.version_info)
+            + "). Please use a newer version of Python."
+        )
```
eddieantonio, line 19: Unfortunately, there is no `require v3.7`; however, this has a similar effect to the Perl idiom; exiting with an error status and no traceback when the version is too old.

joanise: Wow, it's just silly there isn't a briefer way to do this! Thanks for providing the solution!

joanise (approved):

PR #75, opened 2021-09-07
feat(anchors): support anchors elements in alignment

joanise:
When an anchor (e.g., `<xml anchor time="15.3s">`) is added to the XML file to align, split the alignment task around it, thus forcing text before/after to match audio before/after, strictly.

While the do-not-align audio (DNA) feature of the config file is still supported, anchors can also be used to marks parts of the audio that should not be align, in one of three ways:
 - if an anchor with time t precedes the first word, audio time [0, t] is excluded from alignment
 - if two anchors with time t1 and t2 are inserted between two words/sentences/paragraphs/pages, audio time [t1, t2] is excluded from alignment
 - if an anchor with time t is found after the last word, audio time [t, end] is excluded from alignment.

Silence splitting is done in an anchors-aware way: the silence before or after a word, which is split between neighbouring words, will not jump over an anchor or the boundary of a DNA segment.

DNA audio and anchors can both be used in the same alignment specification. One just has to make sure they are coherent and match what's actually happening in the audio, otherwise alignment will probably fail.


joanise:
I've done a careful second review of my own work, refactored a few things and improved documentation, and now I'm ready to merge it it.

PR #76, opened 2021-09-28
Test.toby: refactor 

tobyatgithub:
Adding two refactors, one about parse_g2p_fallback from cli.py, another one about getting languages from views.py


joanise:
If you inspect the build log, you will see that some unit tests are failing: https://app.travis-ci.com/github/ReadAlongs/Studio/builds/238613355
You can run the unit tests on your own machine:
```
cd Studio/test
./run.py dev
```
It's not easy to troubleshoot from there, but I think it all boils down to `LANGS` not being defined in `parse_g2p_fallback`.

joanise:
This PR is unfortunately quite messy. There are a lot of commits that add and remove unnecessary files, and a lot of merges. Before merge the PR in, I'd like it cleaned up to just one or two relevant, stand-alone commits, put on the tip of the current `master` branch.

Also, when refactoring a function like `parse_g2p_fallback()`, it would be best to add a direct test case for it to the test suite. Refactoring code is always safest when done with unit testing, and the refactoring is not complete until the unit tests pass. Here the unit tests caught the problem, but I agree they do not make it easy to understand where the issue is. Adding specific cases for this function will make the output clearer.

joanise:
You can clean up the PR history by doing a squashed merge on a new branch, e.g., something like
```
git checkout master  # make sure this points exactly to origin/master on https://github.com/ReadAlongs/Studio
git checkout -b dev.refactor-langs
git merge --squash test.toby
```
and then work from these to fix the test suites. It might make most sense to create a new PR from that and close this PR, just to have it all clean and tidy.

I generally avoid working on long standing branches in my own forks. When I bring in upstream changes from someone else's project, or even from my own, I tend to rebase my own changes onto the upstream's master and create a clean history for pushing. I have messy forks I work on, but I never push that stuff up, I try to make the history that's shared with the world easy to understand: just the relevant commits, well documented, well tested.

PR #77, opened 2021-09-29
refactor: change pystache to chevron

roedoejet:
Following @tobyatgithub's report that the pystache package is causing problems, this refactors it out of the codebase in favour of [chevron](https://github.com/noahmorrison/chevron).


readalongs/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.20210825"
+__version__ = "0.1.20210928"
```
joanise, line 1: How did this pass `black --check`? CI should have complained about this file. `black --check readalongs/_versino.py` complains on my machine...

readalongs/align.py
```diff
@@ -53,7 +53,7 @@
 
 @dataclass
 class WordSequence:
-    """ Sequence of "unit" XML elements
+    """Sequence of "unit" XML elements
```
joanise, line 56: You're the second person to remove my space between `"""` and the text. I assume your editor did it automatically? When I run black, it doesn't change this, but I'll try to remember no to insert that space in the future.

requirements.txt
```diff
@@ -12,7 +12,7 @@ soundswallower==0.1.1
 pydub==0.23.1
 pympi-ling==1.69
 python-slugify==5.0.0
-pystache==0.5.4
+chevron==0.14.0
```
joanise, line 15: Just being finicky, we're normally keeping this file sorted...

test/data/fra-prepared.xml
```diff
@@ -11,7 +11,7 @@
                     <s>Et voici une deuxième ligne dans le même paragraphe.</s>
                 </p>
                 <p>
-                    <s>Un deuxième paragraphe précédé d&#x27;une ligne vide est ici. Avec deux phrases.</s>
+                    <s>Un deuxième paragraphe précédé d'une ligne vide est ici. Avec deux phrases.</s>
```
joanise, line 14: It took me a while to figure out why you made this change, since it's semantically the same, but I see chevron outputs literal `'` instead of `&#x27;`. It's much nicer this way, so good.

joanise (approved):
Nicely simple in-place switch. Looks good.

readalongs/_version.py
joanise: In case it's not clear, black wants to add a newline at the end of this file.

requirements.txt
roedoejet: ah, no, that's a good point.

readalongs/align.py
roedoejet: Hm. interesting. Yes it was done automatically but I assumed it was black. Maybe it's vscode. I'll pay more attention next time.

PR #78, opened 2021-10-01
Dev.package

roedoejet:


test/data/ej-fra-package.xml
```diff
@@ -0,0 +1,29 @@
...
+            <div type="page">
+                <graphic url="data/diamond.png"/>
+                <p>
+                    <s>Bonjour.</s>
+                    <s>Je m'appelle Éric Joanis.</s>
+                    <s>Je suis programmeur au sein de l'équipe des technologies pour les langues autochtones au CNRC.</s>
+                </p>
+            </div>
+            <div type="page">
+                <graphic url="https://cdn.shibe.online/shibes/bdf0a814c42e39e42870ffb22075ee0a9f1db933.jpg"/>
```
joanise, line 17: I'm still not comfortable with the use of this image in our test suite. Free sites are often only free for personal use, but require a paid license for use by governments and corporations. Can we use something we own, as in the NRC? How about https://nrc.canada.ca/libraries/theme-gcweb/assets/sig-blk-en.svg ?

joanise: Of course, the link might break in the future, so it's maybe not ideal, but if we use something from shibe, I'd like to keep a link to the terms of use right here to prove we're allowed to use it.

test/test_align_cli.py
```diff
@@ -126,6 +126,21 @@ def test_invoke_align(self):
...
+                join(output, "html"),
+                "--html",
+            ],
+        )
+        self.assertEqual(results_html.exit_code, 0)
+        self.assertTrue(
+            exists(join(output, "html", "html.html")),
+            "succesful html alignment should have created output.html",
+        )
+
```
joanise, line 143: Hum, this is pretty weak validation... Maybe inspect the contents and make sure it contains the strings `read-along text="data:application/xml;base64`, `alignment="data:application/smil+xml;base64` and `audio="data:audio/mp4;base64`?

Also, the message in case of error should say "should have created html/html.html", rather than "output.html".

readalongs/cli.py
```diff
@@ -153,6 +152,7 @@ def cli():
 @click.option(
     "-t", "--text-grid", is_flag=True, help="Export to Praat TextGrid & ELAN eaf file"
 )
+@click.option("-H", "--html", is_flag=True, help="Export to WebComponent HTML")
```
joanise, line 155:
```suggestion
@click.option("-H", "--html", is_flag=True, help="Export to a single-file HTML format")
```
"WebComponent HTML" is not very meaningful to me.

joanise (approved):
Great work, thanks! I hope you don't mind the nitpicking too much.

PR #80, opened 2021-10-05
Dev.silence

roedoejet:
This PR combines a few things.

- a small fix to the unittests for the `dev.package` PR (c25502afe90e6589e711aca6975db981cbf2275d)
- a refactor of the way test cases are set up. our previous code was not very dry, so I created a BasicTestCase that can be inherited in any test cases that need access to the temporary directory and application runner. I updated all our tests to run that way. (f1896148398db2d09c3b7b543f3821f7da00240e)
- a feature of inserting silence at specific places (1663779ddfd7f04813184c8325cef48b461596c9). any element can be added in the xml, I used span elements, but these could be anything. The xpath just looks for elements with a `silence` attribute. Then silences (in milliseconds) are inserted at those points in the audio following alignment. The start/end alignments for all other words are adjusted accordingly.


readalongs/align.py
```diff
@@ -413,7 +415,36 @@ def frames_to_time(frames):
...
+    words_dict = {
+        x["id"]: {"start": x["start"], "end": x["end"]} for x in results["words"]
+    }
+    silence_offsets = defaultdict(int)
+    silence = 0
+    if results["tokenized"].xpath("//*[@silence]"):
+        endpoint = 0
+        for el in results["tokenized"].xpath("//*"):
+            if "silence" in el.attrib:
+                silence_ms = int(el.attrib["silence"])  # get silence (ms)
```
joanise, line 427: I have a change request on this one. Can we use `parse_time()` from `readalongs.text.util`? That'll change your semantics, because the default unit will become the second, but anchors are in seconds by default, so I think that's appropriate, important even, for consistency.

Second reason for that ask is that Patrick requested I support the default format Audacity displays, which is `00h00m00.000s`. When I fix `parse_time()` to support that, it would be good for the silence syntax to benefit from the change too.

joanise: By the way, it would be tempting to use an existing library for this, but I'd rather not introduce yet another dependency. It's not a big job to fix `parse_time()`, so I will do it.

joanise (changes requested):
Nice work. Great feature to add, with some changes requested below.

Let's not forget to document this feature, too!

readalongs/align.py
joanise: @roedoejet  Just pushed a change to `parse_time()` on this PR's branch, accepting h/m/s/ms. I was going to do it on its own PR, but then you'd have had to merge it into this one so I figured tacking it on to your PR made sense. I'll appreciate a review of those changes, if you don't mind.

test/utils.py
```diff
@@ -0,0 +1,38 @@
+import os
+import tempfile
+from unittest import TestCase, main
+
+from readalongs.app import app
+from readalongs.log import LOGGER
+
+
+class BasicTestCase(TestCase):
```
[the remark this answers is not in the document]
roedoejet: great! sure, I'll change the name.

readalongs/align.py
roedoejet: great, thanks for the quick addition! and yes, that makes sense to be consistent with the anchors feature.

PR #81, opened 2021-10-07
Shape up our readthedocs documentation

joanise:
Built from Fin's documentation at https://github.com/finguist/ReadAlong-Studio-Documentation.git, this PR tried to restructure all the documentation into a coherent whole that can get published to readthedocs.

I did a fair bit of rewriting and editing and the same time, to improve the overall quality of our documentation.

Once approved, the next step will be to publish the results to readthedocs.

### DIFF
--- a/docs/advanced-use.rst
+++ b/docs/advanced-use.rst
@@ -82,9 +82,14 @@ Troubleshooting
 ---------------
 
 Here are three types of common errors you may encounter when trying to
-run ReadAlongs, and ways to debug them. ### Phones missing in the
-acoustic model You may get an error that looks like this:|image1| The
-general structure of your error would look like
+run ReadAlongs, and ways to debug them.
+
+Phones missing in the acoustic model
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You may get an error that looks like this:|image1|
+
+The general structure of your error would look like
 ``Phone [character] is missing in the acoustic model; word [index] ignored``
 This error is most likely caused not by a bug in your ReadAlong input
 files, but by an error in one of your g2p mappings. The error message is
@@ -110,8 +115,8 @@ Follow these steps to debug the issue **in g2p**.
       message shown above: ``g`` and ``ʰ``.
 
 2. Once you have isolated the characters that are not being converted to
-   eng-arpabet, you are ready to begin debugging the issue. Go through
-   steps 3 - ? for each problematic character.
+   eng-arpabet, you are ready to begin debugging the issue. Start at
+   step 3 below for each problematic character.
 
 3. Our next step is to identify which mapping is converting the
    problematic characters incorrectly. Most of the time, the issue will
--- a/docs/start.rst
+++ b/docs/start.rst
@@ -27,17 +27,17 @@ optionally be generated as an ePub file.
 Required knowledge
 ------------------
 
--  Command line interface (CLI)
--  Plain text file/xml/smil
--  Audacity or similar
--  Spinning up a server
+-  How to use a `Command-line interface (CLI) <https://en.wikipedia.org/wiki/Command-line_interface>`__.
+-  How to edit and manipulate plain text, `XML <https://www.w3.org/standards/xml/core>`__ and `SMIL <https://www.w3.org/TR/smil/>`__ files using a text editor or a code editor.
+-  How to edit and examine an audio file with `Audacity <https://www.audacityteam.org/>`__ or similar software.
+-  How to spin up a local web server (e.g., see `How do you set up a local testing server? <https://developer.mozilla.org/en-US/docs/Learn/Common_questions/set_up_a_local_testing_server>`__)
 
 What you need to make a ReadAlong
 ---------------------------------
 
 In order to create a ReadAlong you will need two files:
 
-- Plain text (``.txt``) or XML (``.xml``)
+- A text file, either in plain text (``.txt``) or in XML (``.xml``)
 - Clear audio in any format supported by `ffmpeg <https://ffmpeg.org/ffmpeg-formats.html>`__
 
 The content of the text file should be a transcription of the audio


docs/start.rst
```diff
@@ -3,7 +3,44 @@
...
+Which can be loaded using the `read-along web
+component <https://github.com/roedoejet/ReadAlong-Web-Component>`__.
+
+A book is generated as a standalone HTML page by default, but can
+optionally be generated as an ePub file.
+
+Required knowledge
+------------------
+
+-  Command line interface (CLI)
```
roedoejet, line 30: This is great, but I would like to see links for each of these. Like maybe a link to https://developer.mozilla.org/en-US/docs/Learn/Common_questions/set_up_a_local_testing_server with the "spinning up a server" item. Also, I think verbs are better than nouns for these lists - ie what exactly do we expect them to do with "Plain text file/xml/smil"? Maybe change to "Edit and manipulate plain text, XML (linked), and SMIL (linked) files using a text or code editor" for example

README.md
```diff
@@ -1,10 +1,11 @@
...
 
 [![codecov](https://codecov.io/gh/ReadAlongs/Studio/branch/master/graph/badge.svg)](https://codecov.io/gh/ReadAlongs/Studio)
-[![Build Status](https://travis-ci.com/ReadAlongs/Studio.svg?branch=master)](https://travis-ci.com/ReadAlongs/Studio)
+[![Build Status](https://travis-ci.com/ReadAlongs/Studio.svg?branch=master)](https://travis-ci.com/github/ReadAlongs/Studio)
 [![PyPI package](https://img.shields.io/pypi/v/readalongs.svg)](https://pypi.org/project/readalongs/)
 [![GitHub license](https://img.shields.io/github/license/ReadAlongs/Studio)](https://github.com/ReadAlongs/Studio/blob/master/LICENSE)
 [![standard-readme compliant](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/ReadAlongs/Studio)
+[![Documentation Status](https://readthedocs.org/projects/readalong-studio/badge/)](https://readalong-studio.readthedocs.io)
 
 :warning: :construction: This repo is currently **under construction** and is not stable. :construction: :warning:
```
roedoejet, line 10: I feel like we can take the "under construction" warning out now. Sure, we're active, but we're not nearly as unstable as we were back then.

docs/advanced-use.rst
```diff
@@ -0,0 +1,157 @@
...
+Manipulating the text and/or audio data that you are trying to align can
+sometimes produce longer, more accurate ReadAlongs, that throw less
+errors when aligning. While some of the most successful techniques we
+have tried are outlined here, you may also need to customize your
+pre-processing to suit your specific data.
+
+Audio pre-processing
+~~~~~~~~~~~~~~~~~~~~
+
+Adding silences
```
roedoejet, line 18: I didn't know we did this... then we have to undo this though in the eventual readalong right? This sounds a bit unnecessary if we have anchors.

docs/advanced-use.rst
```diff
@@ -0,0 +1,157 @@
...
+   soundtrack.export(wfile, format="mp3")
+
+Text pre-processing
+~~~~~~~~~~~~~~~~~~~
+
+Numbers
+^^^^^^^
+
+ReadAlong Studio cannot align numbers written as digits (ex. "123").
+Instead, you will need to write them out (ex. "one two three" or "one
```
roedoejet, line 67: This isn't totally true if G2P has a numbers processing mapping. I think I write about this in the blog so hopefully I remember to link it here as an alternative option.

docs/advanced-use.rst
```diff
@@ -0,0 +1,157 @@
...
+::
+
+   num2words 123456789
+   one hundred and twenty-three million, four hundred and fifty-six thousand, seven hundred and eighty-nine
+
+Troubleshooting
+---------------
+
+Here are three types of common errors you may encounter when trying to
+run ReadAlongs, and ways to debug them. ### Phones missing in the
```
roedoejet, line 85: What's with these ### ?

docs/advanced-use.rst
```diff
@@ -0,0 +1,157 @@
...
+      have isolated the right characters by ensuring every other
+      character in your error message appears as an **output** in the
+      `eng-ipa-to-arpabet
+      mapping <https://github.com/roedoejet/g2p/blob/master/g2p/mappings/langs/eng/eng_ipa_to_arpabet.json>`__.
+      These are the problematic characters we need to debug in the error
+      message shown above: ``g`` and ``ʰ``.
+
+2. Once you have isolated the characters that are not being converted to
+   eng-arpabet, you are ready to begin debugging the issue. Go through
+   steps 3 - ? for each problematic character.
```
roedoejet, line 114: ? should be something else? 8?

docs/advanced-use.rst
```diff
@@ -0,0 +1,157 @@
...
+   that eng-ipa-to-arpabet can read. These should be the same characters
+   you identified in step 2.
+
+7. Run ``g2p convert dog xyz xyz-ipa``. Ensure the result is what you
+   expect. If not, your error may arise from a problem in this mapping.
+   refer_to_g2p_troubleshooting. If the result is what you expect,
+   continue to the next step.
+
+8. Note the result from running the command in 7. Check that the
+   characters [TODO-fix this text] (appear/being mapped by generated --
```
roedoejet, line 154: Hm, I don't really understand what this bullet point is saying.

docs/cli-guide.rst
```diff
@@ -0,0 +1,389 @@
...
+
+So, a full command for a story in Algonquin would be something like:
+
+``readalongs prepare -l alq Studio/story.txt Studio/story.xml``
+
+The generated XML will be parsed in to sentences. At this stage you can
+edit the XML to have any modifications, such as adding ``do-not-align``
+as an attribute of any element in your XML.
+
+.. _dna:
```
roedoejet, line 79: I'm not sure this should be in the CLI section - it's more of a text/xml/markup editing activity (along with anchors and other things. Maybe that should have its own section?

docs/advanced-use.rst
joanise: Fin used it quite a bit, but it's true with the anchors feature this might be obsolete. I think where she did this, she kept the inserted silences in the results.

docs/advanced-use.rst
joanise: But in most languages it doesn't having such a processing mapping. But yes, if you write about it in the blog, a link here would be great.

docs/advanced-use.rst
joanise: Ooops. That was supposed to be a sub-heading, which pandoc failed to convert, which I failed to notice. I'll fix it.

joanise: So, now I see (it's more clear here: https://github.com/finguist/ReadAlong-Studio-Documentation/blob/main/prepare.md) that the troubleshooting section only has one error case covered. I'm sure there were more that got discussed, but this is the only one that got documented.

joanise: Now, the software changed since Shanks wrote this item, so the intro text should be modified. The simpler solution is to use `--g2p-fallback und` and in most cases the problem just goes away without needing to fix any g2p mapping. But I want to keep the information about fixing the g2p mapping, but that's also going to be relevant sometimes.

joanise: In any case, I'll improve this section a bit, but it'll still need work which I'm not up to doing right now.

docs/cli-guide.rst
joanise: Hum, it's here because it's part of the story told in the guide, of why you might want to run prepare. The steps suggested are 1. run prepare 2. mark you DNAs, 3. run align, and the file tries to follow that structure. But you're right, a section with dna, anchors, silences, etc, might make more sense.

PR #82, opened 2021-10-07
feat(package): add packaging of fonts and js bundles

roedoejet:
This updates the dev.package feature to include font and js bundles as generated by the web component. When the web component is published, I will add tests to check for 200 responses for the bundles. This PR will be waiting until https://github.com/ReadAlongs/Web-Component/pull/18 is merged.


joanise (approved):
Looks good to me. I spliced the align --html test into its own function in test_align_cli.py, to make it a bit cleaner.

I like the logic of having a saved copy in this repo that gets used as a fallback.

PR #85, opened 2021-10-18
refactor(silence): change syntax for adding silence and allow output …

roedoejet:
…to variety of audio formats

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -21,9 +21,10 @@ from typing import Dict, List, Union
 import chevron
 import regex as re
 import soundswallower
+from humanfriendly.terminal import output
 from lxml import etree
 from pydub import AudioSegment
-from pydub.exceptions import CouldntEncodeError
+from pydub.exceptions import CouldntDecodeError, CouldntEncodeError, PydubException
 from pympi.Praat import TextGrid
 from webvtt import Caption, WebVTT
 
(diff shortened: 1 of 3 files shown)


readalongs/align.py
```diff
@@ -420,11 +420,11 @@ def frames_to_time(frames):
...
     silence_offsets = defaultdict(int)
     silence = 0
-    if results["tokenized"].xpath("//*[@silence]"):
+    if results["tokenized"].xpath("//silence"):
         endpoint = 0
         for el in results["tokenized"].xpath("//*"):
-            if "silence" in el.attrib:
-                silence_ms = parse_time(el.attrib["silence"])
+            if el.tag == "silence" and "dur" in el.attrib:
+                silence_ms = parse_time(el.attrib["dur"])
```
joanise, line 427: This needs a try/except KeyError to gracefully handle bad inputs. With anchors, I made it a hard error with what I tried to make a clear error message (see `get_sequences()`).

I'm torn between making the silence one a warning or a similarly hard error. Warning would issue a message but ignore the silence request. That's seemingly friendlier, in the spirit of "do something reasonable with the input given", but it would also lead to frustrating troubleshooting, à la "I put a silence in there, why does my read-along not have that silence?", which can be prevented with a hard error instead.

I vote hard error.

readalongs/align.py
```diff
@@ -518,11 +518,13 @@ def save_readalong(
...
 
     _, audio_ext = os.path.splitext(audiofile)
+    audio_path = output_base + audio_ext
+    audio_format = audio_ext[1:]
     if audiosegment:
-        audio_path = output_base + ".wav"
-        audiosegment.export(audio_path, format="wav")
+        if audio_format in ["m4a", "aac"]:
+            audio_format = "ipod"
+        audiosegment.export(audio_path, format=audio_format)
```
joanise, line 526: Boy, it's unfortunate we have to do this. I can imagine so many other cases that could require different values for `format`. I hope we're not opening a tricky can of worms here!

I think it can stay as is for now, but maybe watch out in the future, and choose a default format to use instead? We could also implement a fallback: if this export fails (we'd need to figure out how to test for that!), then fallback to a format known to be generally supported (like AAC, or maybe Ogg? https://www.makeuseof.com/tag/audio-file-format-right-needs/).

joanise: I suspect this is where your unit test is failing on Xenial, which suggests if we want the code to be fairly portable to machines that are not all super recent, it might be worth trying my idea and seeing if that makes the unit tests pass on Xenial.

joanise (commented):
Thanks for changing the syntax format and adding these tests. And sorry they're so finicky!

suggested changes, explained further in the comments:
 - try/catch the call to parse_time
 - use a fallback audio format when writing the file with added silences fails

PR #87, opened 2021-11-01
fix(video): force audio mimetypes for video formats

roedoejet:
This is the small fix I mentioned to fix Marc Tessier's bug. Ideally we should be extracting the audio from video formats though.


joanise (approved):
Looks fine. Hacky but it works, that's what we needed.

PR #90, opened 2021-11-03
Dev.cli

roedoejet:
This is a placeholder PR while @joanise and I work on a series of improvements to the command line interface.

- [x] Replace readalongs -H, -x, -t, -C by -o <format>
- [x] Remove epub
- [x] Remove -u switch from readalongs align
- [x] Improve readalongs -h documentation
- [x] Improve readalongs -l with l1:l2:l3
- [x] Add readalongs langs
- [x] Change readalongs align -i
- [x] Update the documentation to match the CLI changes 

Once the remaining issues are added, we can merge this and make a breaking change release (bump major version number)


readalongs/align.py
```diff
@@ -509,30 +504,41 @@ def save_readalong(  # noqa C901
...
         audio = read_audio_from_file(audiofile)
         duration = audio.frame_count() / audio.frame_rate
         words, sentences = return_words_and_sentences(align_results)
         textgrid = write_to_text_grid(words, sentences, duration)
+
+    if "TextGrid" in output_formats:
         textgrid.to_file(output_base + ".TextGrid")
+
+    if "eaf" in output_formats:
         textgrid.to_eaf().to_file(output_base + ".eaf")
```
joanise, line 518: It's logically identical, but I would find it easier to read if these two if statements were indented under the above if TextGrid or eaf block. It makes it more obvious we're still dealing with that type of output.

ditto the if srt and if vtt below, I'd put them inside the if srt or vtt block.

readalongs/cli.py
```diff
@@ -77,12 +83,58 @@ def get_click_file_name(click_file):
 
 
 @click.version_option(version=__version__, prog_name="readalongs")
-@click.group(cls=FlaskGroup, create_app=create_app, context_settings=CONTEXT_SETTINGS)
+@click.group(context_settings=CONTEXT_SETTINGS)
 def cli():
-    """Management script for Read Along Studio."""
+    """Management script for ReadAlong Studio.
```
joanise, line 88: Nice, that is so much better a global help message. All of it useful. It formats nicely on screen, did you check that it generates a good output in readthedocs? Actually, it doesn't matter, our readthedocs does not embed the output of `readalongs -h`, it gives this information in more details instead.

readalongs/cli.py
```diff
@@ -101,10 +153,12 @@ def cli():
...
-    "-C",
-    "--closed-captioning",
-    is_flag=True,
-    help="Export sentences to WebVTT and SRT files",
+    "-o",
+    "--output-formats",
+    type=click.Choice(SUPPORTED_OUTPUT_FORMATS.keys()),
+    multiple=True,
+    help="The output file formats to export to. By default the text is exported as XML and alignments are exported as SMIL. "
+    + f"Other formats include:\b \n\n{SUPPORTED_OUTPUT_FORMATS_DESC}",
```
joanise, line 161: This is mostly good, but on my terminal `readalongs align -h` add a lot of whilespace here, but a full blank line between each format. That seems excessive to me. I'd prefer something like
```
                                  alignments are exported as SMIL. Other
                                  formats include:
                                   - eaf - ELAN file
                                   - html - Single-file, offline HTML
                                   - srt - SRT subtitle
                                   - TextGrid - Praat TextGrid
                                   - vtt - WebVTT subtitle
                                   - xhtml - Simple XHTML
```

readalongs/cli.py
```diff
@@ -267,43 +306,21 @@ def align(**kwargs):  # noqa: C901
...
         output_dir=output_dir,
         output_basename=output_basename,
         config=config,
-        text_grid=kwargs["text_grid"],
-        closed_captioning=kwargs["closed_captioning"],
-        output_xhtml=kwargs["output_xhtml"],
         audiofile=kwargs["audiofile"],
         audiosegment=results["audio"],
-        html=kwargs["html"],
+        output_formats=output_formats,
```
joanise, line 319: So much tidier. :)

readalongs/cli.py
```diff
@@ -29,13 +24,11 @@
 from tempfile import TemporaryFile
 
 import click
-from flask.cli import FlaskGroup
```
joanise, line 32: Does this change mean we're actually no longer supporting your old web page app at all anymore? Or is using @cli.command vs @app.cli.command still compatible with views.py?

test/test_align_cli.py
```diff
@@ -28,8 +28,14 @@ def test_invoke_align(self):
...
-                "-C",
-                "-t",
+                "-o",
+                "vtt",
+                "-o",
+                "srt",
+                "-o",
+                "TextGrid",
+                "-o",
+                "eaf",
```
joanise, line 38: Do we also support `-o vtt:srt:TextGrid:eaf`? I would prefer having that option when I'm using these interactively, it seems to me Click can support that. I don't know if `:` is the default separator. If click supports it, I might use the same thing for specifying multiple languages via `-l`.

readalongs/cli.py
joanise: No, just tested, it does not. Nice to remove unneeded imports for sure, especially when they only add stuff we never use (anymore, anyway).

joanise (commented):
Took the time to review the stuff so far.

readalongs/align.py
roedoejet: sure, either way doesn't bug me.

readalongs/cli.py
roedoejet: good point, I'll try and take away that extra white space

readalongs/cli.py
roedoejet: you can always run it with `python3 run.py` - I'll remove the app import and create_app function too

test/test_align_cli.py
roedoejet: no I don't believe that's supported unfortunately. I'm not sure click supports that either...it seems odd. I see how you can do [fixed width options](https://click.palletsprojects.com/en/8.0.x/options/#:~:text=message%20and%20exit.-,Multi%20Value%20Options,-%C2%B6), but that's not what we want either.

readalongs/cli.py
roedoejet: hm. if I do just a single \n it produces this 
image
which is worse. ideas?

roedoejet: image
 with pipes?

joanise: Darn, I just tested and \b doesn't solve it either. How about `eaf (ELAN file), html (Single-file, offline HTML), srt (SRT subtitle), ...`

test/test_align_cli.py
joanise: Well, it's quite easy to implement the way I did g2p_fallback, with a `.split(":")`.

joanise: Another option, we can do it with a callback:
```
def joiner_callback(valid_values, joiner=":"):
    def _callback(ctx, param, value):
        results = [x for v in value for x in v.split(joiner)]
        for x in results:
            if x not in valid_values:
                raise click.BadParameter(f"'{x}' is not one of {', '.join(valid_values)}.")
        return results
    return _callback
...
@click.option(
    "-o",
    "--output-formats",
    multiple=True,
    callback=joiner_callback(SUPPORTED_OUTPUT_FORMATS),
    ...
```
And I'll be able to reuse it for `-l` too, with `callback=joiner_callback(LANGS)`.

I'm almost done implementing this change, so I'll clean it up and commit it.

joanise: changes pushed.

readalongs/align.py
joanise: change done and pushed.

readalongs/cli.py
roedoejet: sure, that would work

roedoejet: done

docs/cli-guide.rst
```diff
@@ -49,28 +48,31 @@ The plain text file must be plain text encoded in ``UTF-8`` with one
...
 
 The languages supported by RAS can be listed by running ``readalongs prepare -h``
 and they can also be found in the :ref:`cli-prepare` reference.
 
-So, a full command for a story in Algonquin would be something like:
+So, a full command for a story in Algonquin, with a g2p fallback to
+Undetermined, would be something like:
 
-``readalongs prepare -l alq Studio/story.txt Studio/story.xml``
+``readalongs prepare -l alq,und Studio/story.txt Studio/story.xml``
```
roedoejet, line 75: Although, if we are changing `und` to default anyways, this is redundant.

docs/cli-guide.rst
```diff
@@ -223,76 +229,96 @@ separate elements in a list or dictionnary, but if you accidentally have
...
-To enable the g2p cascade, add the ``--g2p-fallback l1:l2:...`` option
-to ``readalongs g2p`` or ``readalongs align``:
+To enable the g2p cascade, provide multiple languages via the ``-l`` switch
+(for plain text input) or add the ``fallback-langs="l2,l3,...`` attribute to
+any element in the XML file:
 
-::
+.. code-block:: xml
+
+   <s xml:lang="eng" fallback-langs="fra,und">English mixed with français.</s>
```
roedoejet, line 275: Again, with the explicit `und`

docs/cli-guide.rst
```diff
@@ -223,76 +229,96 @@ separate elements in a list or dictionnary, but if you accidentally have
...
+   readalongs align -l fra,eng,und myfile.txt myfile.wav output-dir
 
 The "Undetermined" language code: und
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Notice that the two examples above use ``und`` as the last language in the
 cascade. ``und``, for Undetermined, is a special language mapping that
-uses the Unicode definition of all known characters in all alphabets, and
+uses the definition of all characters in all alphabets that are part of the
+Unicode standard, and
```
roedoejet, line 291: Add sentence about the default

docs/cli-guide.rst
joanise: Agreed, it will be. I guess I'll have to review the documentation yet again when I work on that change.

readalongs/python_version.py
```diff
@@ -1,27 +0,0 @@
-"""
-Specifies the minimum version of Python required for this project.
```
roedoejet, line 2: Why have we removed all this in this PR?

test/test_align_cli.py
```diff
@@ -357,6 +367,100 @@ def test_misc_align_errors(self):
...
+            )
+        self.assertNotEqual(results.exit_code, 0)
+        self.assertIn("Error parsing XML", results.output)
+
+        # Giving -i switch generates an obsolete-switch error message
+        with SoundSwallowerStub("word:0:1"):
+            results = self.runner.invoke(
+                align,
+                [
+                    "-i",
```
roedoejet, line 455: Nice testing throughout here.

roedoejet (commented):
Looks good to me! I can't approve or do anything like that. But, I just made a few comments (the biggest change being unfortunately going through the docs again w.r.t the "und" fallback. I think this is ready!

readalongs/python_version.py
joanise: Well, it's true it's not related to this PR, but it's code clean up. I moved that test to `__init__.py` and made it much simpler. The module init file is really the right place, since that's the one file we know gets loaded no matter how you enter into the module, and while I'm grateful for the solution Eddie implemented, I think it was much too verbose, so I refactored it.

docs/cli-guide.rst
joanise: done

docs/cli-guide.rst
joanise: done

docs/cli-guide.rst
joanise: rewrote parts of this paragraph to reflect the change in default behaviour.

joanise:
@roedoejet I say let's merge this puppy! I just committed the changes to add `und` by default to languages.

PR #91, opened 2021-11-04
Bug fix: align -i without -l should not output stack trace

### LINKED ISSUE #88: readalongs -i without -l gives stack trace; ditto readalongs with noise-only output
Reproduce:
```
$ readalongs align -i data/ej-fra.txt data/ej-fra.m4a delme -f
...
  File "c:\users\joanise\ras\studio\readalongs\align.py", line 387, in align_audio
    final_end = end
UnboundLocalError: local variable 'end' referenced before assignment
```
Should output CLI error: "missing -l option".

Seems related:
```
$ readalongs align data/ej-fra.xml data/noise_at_1500.mp3 delme -f
...
   File "c:\users\joanise\ras\studio\readalongs\align.py", line 387, in align_audio
    final_end = end
UnboundLocalError: local variable 'end' referenced before assignment
```

Should output: "no non-noise segments founds" or something like it.

joanise:
align that outputs no non-noise segments should also not output stack trace 

To unit test this, I had to add more calls to `align()`, which ended up calling the SoundSwallower decoder just to exercise code in `align.py` itself, so I create a stub that lets us do quick align tests when we don't care about (or can fake) SS's output.

Fixes #88

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -304,6 +304,7 @@ def align_audio(  # noqa: C901
 
     # Extract the list of sequences of words in the XML
     word_sequences = get_sequences(xml, xml_path, unit=unit)
+    end = 0
     for i, word_sequence in enumerate(word_sequences):
 
         i_suffix = "" if i == 0 else "." + str(i + 1)
--- a/readalongs/cli.py
+++ b/readalongs/cli.py
@@ -82,7 +82,7 @@ def cli():
     """Management script for Read Along Studio."""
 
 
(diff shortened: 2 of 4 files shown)


joanise:
PS: This is ready to merge into master before dev.cli gets completed.

roedoejet (approved):
Huh, I've never run into this, but it makes sense! Thanks for the nice quick, easy fix and accompanying unittests here @joanise !

PR #92, opened 2021-11-05
refactor(cli): replace --g2p-fallback option by -l with multiple languages

joanise:
as agreed on Nov 1, 2021 at the RA-S dev meeting, it's much more intuitive to
specify multiple languages with -l that having a separate option. To boot,
using an attribute mean the fallback languages can also be specified on any XML
node.

@roedoejet this CLI feature touches a lot of stuff, so I'd like its own independent review. I'll fast-forward merge it into `dev.cli` once you approve it.

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -171,7 +171,6 @@ def align_audio(  # noqa: C901
     bare=False,
     config=None,
     save_temps=None,
-    g2p_fallbacks=None,
     verbose_g2p_warnings=False,
 ):
     """Align an XML input file to an audio file.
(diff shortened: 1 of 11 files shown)


joanise:
forced push: rebased at the tip of dev.cli and resolved conflict.

roedoejet (commented):
Looks good to me! Even though I doubt many people were using g2p-fallback, I like putting the OBSOLETE help message in there as something we should follow when introducing breaking changes.

joanise:
@roedoejet Yeah, I'm extremely averse to making breaking changes anywhere...

PR #94, opened 2021-11-23
perf: optimize the CLI, mostly by deferring expensive imports

joanise:
This PR speeds up readalongs, mostly by deferring expensive imports.

While there, added some docstrings and did additional relevant refactorings.

More review fun, @roedoejet, in case you didn't already have enough... :)

This PR speeds up the help messages quite a bit by deferring expensive imports to run them only when they're needed. 
Part of the code are cleaner too, but there's also some added complexity in other parts. I think it's a worthwhile trade off, I hope you agree!

### DIFF
--- a/readalongs/cli.py
+++ b/readalongs/cli.py
@@ -13,7 +13,6 @@ CLI commands implemented in this file:
 import io
 import json
 import os
-import re
 import sys
 from tempfile import TemporaryFile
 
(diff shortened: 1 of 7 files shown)


readalongs/util.py
```diff
@@ -1,10 +1,22 @@
...
 LANG_NAMES = None
 
 
+def getLangsDeferred() -> Iterable:
+    """Lazilly get the list of language codes supported by g2p library
+
+    Yields an Iterable in such a way that the g2p database is only loaded when
+    the results are iterated over, rather than when this function is called.
+    """
+    yield from getLangs()[0]
```
roedoejet, line 17: Oh, this is cool!

roedoejet (approved):

readalongs/cli.py
```diff
@@ -48,6 +44,9 @@
 
 def create_app():
     """Returns the app"""
+    # defer expensive import to do it only if and when it's actually needed
+    from readalongs.app import app
```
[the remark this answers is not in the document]
joanise: I hadn't noticed, but yes, that's true, the `INFO - Server initialized for eventlet.` message is now gone, even when you run `readalongs align`. I always wondered what it was from...

readalongs/util.py
joanise: Yeah, generators in Python are strictly lazily evaluated. I tested, and if there's a yield statement in the function, nothing at all in the function runs until `next()` is first called on it.

PR #96, opened 2022-01-07
Better error reporting in test_audio.py

dhdaines:
It seems these errors occur fairly often - it may be more useful to the user to report the full error message.  Also in order to fix dependencies it's necessary to run pip with some extra flags to reinstall.


joanise (approved):
What an excellent idea. These errors bug me every time too, it'll be nice to have a message documenting the known solution when it happens.

PR #100, opened 2022-02-08
docs: embed the list of languages right in the doc

joanise:
Added a call to `readalongs langs` to embed its output right into the docs, for easiest access.

Is it reasonable to add this dependency on https://pypi.org/project/sphinxcontrib-programoutput/ / https://github.com/NextThought/sphinxcontrib-programoutput ? It seems well maintained.

### DIFF
--- a/docs/advanced-use.rst
+++ b/docs/advanced-use.rst
@@ -3,6 +3,8 @@
 Advanced topics
 ===============
 
+.. _adding-a-lang:
+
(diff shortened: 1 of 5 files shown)


roedoejet (approved):
Looks good to me. It's not the most popular package and so we might have some issues if it stops being maintained, but it's an elegant way of documenting the list of languages. Thanks @joanise !

PR #101, opened 2022-03-07
Create an API for ReadAlongs/Studio from the click app.

joanise:
Immediate use case: ReadAlongsDesktop can now avoid repeating code already written by readalongs.cli, by calling the wrappers in readalongs.api.

The advantage of these wrappers over calling the underlying functions in readalongs.align etc is that all the error checking logic implemented in readalongs.cli is applied. The calls are equivalent to command line calls, but without going through a system call.

### DIFF
--- a/readalongs/api.py
+++ b/readalongs/api.py
@@ -9,10 +9,10 @@ def align(textfile, audiofile, output_base, language=None, **kwargs):
     """Run the "readalongs align" command from within a Python script.
 
     Args:
-        textfile (str): input text file (XML or plain text)
-        audiofile (str): input audio file (format supported by ffmpeg)
-        output_base (str): basename for output files
-        language (list[str]): Specify only of textfile is plain text;
+        textfile (str | Path): input text file (XML or plain text)
+        audiofile (str | Path): input audio file (format supported by ffmpeg)
+        output_base (str | Path): basename for output files
+        language (List[str]): Specify only of textfile is plain text;
             list of languages for g2p and g2p cascade
         save_temps (bool): Optional; whether to save temporary files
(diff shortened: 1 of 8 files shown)


readalongs/api.py
```diff
@@ -0,0 +1,146 @@
+"""
+api.py: API for calling readalongs CLI commands programmatically
+
+In this API, functions take the same arguments as on the readalongs
+command-line interface. The mapping between CLI options and API options is
+that the first long variant of an option described in "readalongs cmd -h" is
```
roedoejet, line 6: maybe "readalongs <cmd> -h" or some way to specify that cmd is not an actual command?

readalongs/api.py
```diff
@@ -0,0 +1,146 @@
...
+Example from readalongs align -h:
+    option in CLI                       option in API
+    ================================    =================================
+    -l, --language, --languages TEXT    language=["l1", "l2"]
+    -f, --force-overwrite               force_overwrite=True
+    -c, --config PATH                   config=os.path.join("some", "path", "config.json")
+                                     OR config=pathlib.Path("/some/path/config.json")
+
+As shown above, file names can be constructed using os.path.join() or a Path
+class like pathlib.Path. Warning: don't just use "/some/path/config.json"
```
roedoejet, line 18: I like this - you don't usually see this level of documentation - it's just assumed to be understood, but given that we might be dealing with novice programmers, this is helpful.

roedoejet (approved):
This is a very well-documented and motivated PR that unifies command line and programmatic use of ReadAlongs. Thanks @joanise - this is great.

PR #104, opened 2022-05-24
Update soundswallower to 0.2

dhdaines:
To fix long inputs (such as in ras-evaluation).


requirements.txt
```diff
@@ -8,7 +8,7 @@ g2p>=0.5.20210825
 lxml==4.6.5
 networkx==2.5
 numpy>=1.16.4
-soundswallower==0.1.5
+soundswallower~=0.2
```
joanise, line 11: What does this `~=` syntax mean?

dhdaines: Hi! It means "compatible with" - so 0.2.1 will also be accepted, presuming that I practice proper semantic versioning and all 0.2.x are backwards-compatible with 0.2

dhdaines: Ahh.  I admit not having read the specification very closely, thanks for your attention to this!  I would prefer to restrict to 0.2.* so looks like ~=0.2.0 is the right one to use.

dhdaines:
Should be all ready to go now!  Coverage problems have been fixed by adding an extra test to test_force_align.

joanise:
Starting to review and test this PR in its current state.

joanise:
I just finished rerunning ras-evaluation locally with this PR (having merged in `dev.dummy` locally), and almost all the evaluation numbers go up, except for some of the Flood ones, which was expected. I think this PR is ready to merge in now. Thanks for all that work!

LICENSE
```diff
@@ -2,6 +2,7 @@ MIT License
 
 Copyright (c) 2019 David Huggins-Daines
 Copyright (c) 2019-2021 National Research Council Canada
+Acoustic model Copyright (c) 2015 Alpha Cephei Inc.
```
joanise, line 5:
```suggestion
Acoustic model in readalongs/static/model/cmusphinx-en-us-5.2 Copyright (c) 2015 Alpha Cephei Inc.
```

readalongs/text/make_smil.py
```diff
@@ -35,31 +33,6 @@
 END_SUBIDX = 3
 
 
-def parse_hypseg(text):
```
joanise, line 38: It's nice to have this obsolete code be gone, thanks!

requirements.txt
```diff
@@ -8,7 +8,7 @@ g2p>=0.5.20210825
 lxml==4.6.5
 networkx==2.5
 numpy>=1.16.4
-soundswallower==0.1.5
+soundswallower>=0.2.2
```
joanise, line 11: I'm surprised you didn't use `~=0.2.2`, but this is fine too.

test/test_force_align.py
```diff
@@ -52,6 +55,35 @@ def test_align_text(self):
         for w, xw in zip(words, xml_words):
             self.assertEqual(xw.attrib["id"], w["id"])
 
+    def test_align_switch_am(self):
```
joanise, line 58: Thanks for increasing coverage. It always reassures me when stuff is better tested.

joanise (approved):
Trivial change suggested on the copyright statement, otherwise looks good, ready to merge after that.

requirements.txt
dhdaines: Hmm in fact yes it should be ~= 0.2.2 now that I think of it...

PR #107, opened 2022-06-06
style: move to black 22.3 and reformat everything accordingly

joanise:
We were still using a beta version of black, 19.10b0. With the first
non-beta release, 22.1 (see https://black.readthedocs.io/en/latest/change_log.html),
black now has a stability policy (https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy)
promising that no formatting changes will happen between versions
starting with the same year. By using black~=22.0, we're declaring use
the formatting rules of the 2022 version of black, picking up the latest
in that series of releases.

Changes from the 19.10b0 to 22.3 that are visible in our code and
included in this commit:

 - The trailing comma is no longer allowed when a list fits on one line,
   it's only used when a list is broken on multiple lines, in which
   case it is required.
   So this is OK:

       foo = bar(
           arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, arg11
       )

   But this will get reformatted to one arg per line becasue of the
   trailing comma:

       foo = bar(
           arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, arg11,
       )

   Similarly, this will get reformatted to one arg per line because of
   the trailing comma:

       foo = bar(arg1, arg2,)

 - Docstrings no longer allow extra spaces.
   They are now strictly formatted like this:

       """Short function description

       Long description

       Returns:
           something
       """

   or

       """Short function description"""

   These will now get reformatted like above:

       """ Short function description

           Long description

           Returns:
               something
       """

   or

       """ Short function description """

   Version 19.10b0 of black would have accepted either and left them
   unchanged, but not 22.x.

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -178,8 +178,7 @@ def parse_and_prepare_input(
         lxml.etree.ElementTree: Parsed and prepared XML
 
     Raises:
-        RuntimeError: If XML failed to parse
-"""
+        RuntimeError: If XML failed to parse"""
     # First do G2P
     try:
         xml = etree.parse(xml_path).getroot()
--- a/readalongs/audio_utils.py
+++ b/readalongs/audio_utils.py
@@ -12,8 +12,7 @@ from readalongs.log import LOGGER
 
 
 def join_section(audio: AudioSegment, audio_to_insert: AudioSegment, start: int):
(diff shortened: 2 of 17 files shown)


dhdaines (approved):
All these changes seem quite sensible to me, looks like the new version of Black is really an improvement.

PR #110, opened 2022-06-17
feat: clean up log handling and fix debug_aligner

dhdaines:
Turns out that since -loglevel was implemented, debug_aligner has the unintended side effect of turning *off* most logging.  This changes the handling of logs to be more logical.  We don't send them to /dev/null now that we can control the log level (only errors are printed by default).  With save_temps we turn most logging on and send it to a file.  With debug_aligner we dump everything to the console.

In practice INFO and DEBUG are the same thing but that might not always be the case.

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -236,18 +236,21 @@ def create_asr_config(
     asr_config.set_float("-pbeam", 1e-100)
     asr_config.set_float("-wbeam", 1e-80)
 
-    if not debug_aligner:
-        # With --debug-aligner, we display the SoundSwallower logs on screen, but
-        # otherwise we redirect them away from the terminal.
+    if debug_aligner:
+        # With --debug-aligner, we display the SoundSwallower logs on
+        # screen and set them to maximum strength
+        asr_config.set_string("-loglevel", "DEBUG")
+    else:
+        # Otherwise, we enable logging and direct it to a file if
+        # saving temporary files
         if save_temps is not None and (sys.platform not in ("win32", "cygwin")):
             # With --save-temps, we save the SoundSwallower logs to a file.
             # This is buggy on Windows, so we don't do it on Windows variants
+            # (NOTE: should be fixed in SoundSwallower 0.3 though)
             ss_log = save_temps + ".soundswallower.log"
-        else:
-            # Otherwise, we send the SoundSwallower logs to the bit bucket.
-            ss_log = os.devnull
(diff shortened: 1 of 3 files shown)


dhdaines:
Hmm, looks like I should take the time to ensure code coverage here while I'm at it...  Will do that tomorrow!

joanise:
Thanks for catching this, I noticed it earlier too. I started trying to fix it too, but something is bugging me: I cannot capture SS's log stream (stderr) via our unit testing's `invoke` infrastructure.

I wanted to create a unit test case with `--debug-aligner` and one without, with these assertions:

        # make sure ss logs are disabled:
        self.assertNotIn("Current configuration", results.output)

vs

        # make sure ss logs are enabled:
        self.assertIn("Current configuration", results.output)

except that since the SS logs don't get capture, the `self.assertIn` variant always fails.

Any change you can imagine a way to detect whether `--debug-aligner` is doing what it's supposed to do? And a way to suppress those ss logs from unit testing's on-screen output?

dhdaines:
> Any change you can imagine a way to detect whether `--debug-aligner` is doing what it's supposed to do? And a way to suppress those ss logs from unit testing's on-screen output?

I was contemplating this last night too ... probably the only way to do it is to launch another Python interpreter in a subprocess and capture its output.  It doesn't have to actually do anything useful, as long as it produces some logging output.

SoundSwallower 0.3 will have the ability to redirect SoundSwallower logging through the Python logging module, though I'm not sure if it will do this by default or not.  In this case one would simply add a logging handler and verify that it gets called.

joanise:
Well, then maybe we should just exercise the switch, for coverage's sake, but not bother asserting anything about the output. I'm not sure the subprocess is worth it, although I think we actually do have at least one test that's done that way, so maybe we can use that one here. I'll take a look.

joanise:
Yeah, `test_audio.py` uses subprocess.run

joanise:
I'll commit it onto your branch.

dhdaines:
Ah, thank you! Strange that codecov still complains...

joanise:
Because of the subprocess call, I bet codecov does not have visibility on the fact that it's exercised.

joanise:
Thanks for this patch! It'll be nice with 0.3 when it can be funnelled though Python logging.

I'll approve and merge this in a moment.

joanise (approved):

PR #112, opened 2022-06-23
fix: extract sentences correctly on page changes

### LINKED ISSUE #70: Pratt and VTT output seem to use wrong sentence segmentation
When I run this command in `test/`:

     readalongs align -l fra -s -f -i data/ej-fra.txt data/ej-fra.m4a delme -t  -C

The `.xml` file is fine, but the `.eaf`, `.TextGrid`, and `_sentences.*` files generated consider each sentence on the first page as sentences, but each **word** on the second page as an individual sentence.

The correct output should logically consider the same units as sentences in the `.xml` file and the TextGrid and cc/subtitle files.

joanise:
Identifying the sentence ID of words past the first paragraph was
incorrect, resulting in wrong sentence layers for the vtt, srt and
TextGrid output formats.

The sentence id needs to be the whole word's ID up to be excluding the
word ID itself. I.e. d0p0s0w0 and d0p1s0w0 are not part of the same
sentence, since they're in the first sentence of different paragraphs.

while I'm in here, refactor a bit and clean up documentation.

And include unit testing.

FIXES #70


joanise:
Discussed at the last Dev meeting: IDs should not be parsed at all. Relying on the pattern `(.*\d)w\d+` to identify the sentence is still actually wrong, because we say a use can pre-assign IDs themselves and if they do, we'll just use them as provided.

So, the correct solution we agreed to is to look at the XML: `<w>` elements that share the same `<s>` ancestor are in the same sentence. Edge case: a `<w>` without a `<s>` ancestor is considered a sentence by itself.

joanise:
@roedoejet This work is finally complete and ready to review.

 - We now extract sentence correctly based on the XML structure itself.
 - The text of the words is now collected from all sub-elements via `get_word_text()` everywhere we need that text.

roedoejet (approved):
This looks good to me, thanks @joanise !

PR #115, opened 2022-07-09
feat: Add header and theme from config.json

dhdaines:


docs/cli-guide.rst
```diff
@@ -196,13 +196,23 @@ Here is that list at the time of compiling this documentation:
...
+currently accepts a few components: adding titles and headers, adding
+images to your ReadAlongs, and DNA audio (see :ref:`dna`).
+
+To add a title and headers to the output HTML, you can use the keys
+`"title"`, `"header"`, and `"subheader"`, for example::
+
+  {
+    "title": "My awesome read-along",
+    "header": "A story in my language",
+    "subheader" "Read by me"
```
joanise, line 214:
```suggestion
    "subheader": "Read by me"
```

readalongs/align.py
```diff
@@ -911,7 +911,15 @@ def save_readalong(
...
     if "html" in output_formats:
         html_out_path = output_base + ".html"
-        html_out = create_web_component_html(tokenized_xml_path, smil_path, audio_path)
+        html_out = create_web_component_html(
+            tokenized_xml_path,
+            smil_path,
+            audio_path,
+            config.get("title", "Title goes here"),
+            config.get("header", "Header goes here"),
+            config.get("subheader", "Subheader goes here"),
```
joanise, line 920: I might put in default title/header/subheader that's going to be suitable as a title/header/subheader for most RAs, but I'm not sure what such good default might be.

Oh wait! Those were already the defaults we were generating, you just moved them out to the config! Well, usable defaults would have value, but they're certainly not required for this PR.

joanise: More importantly, it would be good to use these in `text.save_minimal_index_html()`, by the time we accept them as parameters in the config. Is there anywhere else this information should get plugged in?

joanise (changes requested):
Excellent idea, thanks for doing this! We should do it in the default generated minimal index.html too, but otherwise this is good.

joanise:
@dhdaines Thanks for implementing this patch, and sorry it took me so long to merge it in. I've expanded the concept to the basic HTML and added some testing, and merged the results in.

PR #118, opened 2022-09-23
feat: add /convert_alignment endpoint

joanise:
Feature implementation: the /convert_alignment endpoint support converting alignments to TextGrid, eaf, srt, vtt.

Breaking change: the /langs endpoint's return value has changed in order to be better documented

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -910,7 +910,7 @@ def save_readalong(
     smil = make_smil(
         os.path.basename(tokenized_xml_path),
         os.path.basename(audio_path),
-        align_results,
+        align_results["words"],
     )
     save_txt(smil_path, smil)
 
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -716,24 +716,30 @@ def align_audio(
     return results
 
 
+def get_audio_duration(audiofile: str) -> float:
+    """Return the duration of audiofile in seconds"""
+    audio = read_audio_from_file(audiofile)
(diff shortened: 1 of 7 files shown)


joanise:
Questions for review...

1) The `/convert_alignment` endpoint accepts the output format of part of the JSON input dict. It's probably more "conventional" to do `/convert_to/TextGrid` and `/convert_to/eaf` etc, but a) I don't really know how to do that, and b) since a bunch of parameters already have to come through that JSON structure, I'm not sure I see value in changing it.

2) My `ConvertResponse` structure feels clunky with 4 fields, only two of which are used for TextGrid and eaf, and all four for srt and vtt, but I was trying to unify the four possible conversion output into one response object. Is that OK, or should I try to do better? Can one endpoint return one of several different responses, the way it can accept one of several different requests? (I mean, it's pretty obvious it can, but how to do you do cleanly and keep it user intuitive and well documented?) (Maybe a list of filenames and a list of filecontents? Or a list of (filename, filecontents) pairs?)

joanise:
Another question for review:

3. I felt the need to modify the `/langs` endpoint because I wanted it to document its return value in a clearer way using a `Field` with an `example`. But I couldn't do that without changing the return value of `/langs`, so it's a breaking change and we'll have to change the Studio-Web accordingly.

joanise:
Never mind this PR, I'll close it and submit a new one that doesn't change /langs. Instead, I'll just add documentation to /langs describing the structure of its return value.

PR #119, opened 2022-09-26
feat: add /convert_alignment endpoint

joanise:
Feature implementation: the `/convert_alignment` endpoint supports converting alignments to TextGrid, eaf, srt, vtt.

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -737,6 +737,10 @@ def save_label_files(
         duration: length of the audio in seconds
         output_base (str): Base path for output files
         output_formats (Iterable[str]): List of output formats
+
+    Raises:
+        IndexError: words and tokenized_xml have inconsistent IDs
+        Exception: TODO, not sure what else this can raise
     """
     words_with_text, sentences = get_word_texts_and_sentences(words, tokenized_xml)
     textgrid = create_text_grid(words_with_text, sentences, duration)
@@ -761,6 +765,10 @@ def save_subtitles(
         tokenized_xml: tokenized or g2p'd parsed XML object
         output_base (str): Base path for output files
         output_formats (Iterable[str]): List of output formats
+
+    Raises:
+        IndexError: words and tokenized_xml have inconsistent IDs
+        Exception: TODO, not sure what else this can raise
     """
     words_with_text, sentences = get_word_texts_and_sentences(words, tokenized_xml)
     cc_sentences = write_to_subtitles(sentences)
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -910,7 +910,7 @@ def save_readalong(
     smil = make_smil(
         os.path.basename(tokenized_xml_path),
         os.path.basename(audio_path),
-        align_results,
+        align_results["words"],
     )
     save_txt(smil_path, smil)
 
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -716,24 +716,30 @@ def align_audio(
     return results
 
 
+def get_audio_duration(audiofile: str) -> float:
+    """Return the duration of audiofile in seconds"""
+    audio = read_audio_from_file(audiofile)
+    return audio.frame_count() / audio.frame_rate
+
+
 def save_label_files(
-    align_results: Dict[str, List],
-    audiofile: str,
+    words: List[dict],
+    tokenized_xml: etree.ElementTree,
+    duration: float,
     output_base: str,
     output_formats: Iterable[str],
 ):
     """Save label (TextGrid and/or EAF) files.
 
     Args:
-        align_results (Dict[str, List]): results of alignment
-        audiofile (str): Path to input audio
+        words: list of words with "id", "start" and "end"
+        tokenized_xml: tokenized or g2p'd parsed XML object
+        duration: length of the audio in seconds
         output_base (str): Base path for output files
         output_formats (Iterable[str]): List of output formats
     """
-    audio = read_audio_from_file(audiofile)
-    duration = audio.frame_count() / audio.frame_rate
-    words, sentences = get_words_and_sentences(align_results)
-    textgrid = write_to_text_grid(words, sentences, duration)
+    textual_words, sentences = get_textual_words_and_sentences(words, tokenized_xml)
+    textgrid = create_text_grid(textual_words, sentences, duration)
 
     if "textgrid" in output_formats:
         textgrid.to_file(output_base + ".TextGrid")
@@ -743,18 +749,22 @@ def save_label_files(
 
 
 def save_subtitles(
-    align_results: Dict[str, List], output_base: str, output_formats=Iterable[str]
+    words: List[dict],
+    tokenized_xml: etree.ElementTree,
+    output_base: str,
+    output_formats=Iterable[str],
 ):
     """Save subtitle (SRT and/or VTT) files.
 
     Args:
-        align_results (Dict[str, List]): results of alignment
+        words: list of words with "id", "start" and "end"
+        tokenized_xml: tokenized or g2p'd parsed XML object
         output_base (str): Base path for output files
         output_formats (Iterable[str]): List of output formats
     """
-    words, sentences = get_words_and_sentences(align_results)
+    textual_words, sentences = get_textual_words_and_sentences(words, tokenized_xml)
     cc_sentences = write_to_subtitles(sentences)
-    cc_words = write_to_subtitles(words)
+    cc_words = write_to_subtitles(textual_words)
 
     if "srt" in output_formats:
         cc_sentences.save_as_srt(output_base + "_sentences.srt")
@@ -880,8 +890,9 @@ def save_readalong(
     # Create textgrid object if outputting to TextGrid or eaf
     if "textgrid" in output_formats or "eaf" in output_formats:
         save_label_files(
-            align_results=align_results,
-            audiofile=audiofile,
+            words=align_results["words"],
+            tokenized_xml=align_results["tokenized"],
+            duration=get_audio_duration(audiofile),
             output_base=output_base,
             output_formats=output_formats,
         )
@@ -889,7 +900,8 @@ def save_readalong(
     # Create webvtt object if outputting to vtt or srt
     if "srt" in output_formats or "vtt" in output_formats:
         save_subtitles(
-            align_results=align_results,
+            words=align_results["words"],
+            tokenized_xml=align_results["tokenized"],
             output_base=output_base,
             output_formats=output_formats,
         )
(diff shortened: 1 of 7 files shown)


joanise:
Questions for review...

1) The `/convert_alignment` endpoint accepts the output format of part of the JSON input dict. It's probably more "conventional" to do `/convert_to/TextGrid` and `/convert_to/eaf` etc, but a) I don't really know how to do that, and b) since a bunch of parameters already have to come through that JSON structure, I'm not sure I see value in changing it.

2) My `ConvertResponse` structure feels clunky with 4 fields, only two of which are used for TextGrid and eaf, and all four for srt and vtt, but I was trying to unify the four possible conversion outputs into one response object. Is that OK, or should I try to do better? Can one endpoint return one of several different responses, the way it can accept one of several different requests? (I mean, it's pretty obvious it can, but how to do you do cleanly and keep it user intuitive and well documented?) (Maybe a list of filenames and a list of filecontents? Or a list of (filename, filecontents) pairs?)

joanise:
Argh, CI is failing once again because of a failure to push to codecov. This problem usually goes away by itself after a few minutes, but it's a nuisance. The tests passed on my latest push, though.

readalongs/text/make_smil.py
```diff
@@ -28,17 +30,72 @@
...
+        "end": word_end_time_in_seconds (float),
+    }
+
     Args:
-        text_path(str): path to text
-        audio_path(str): path to audio
-        results(dict): all alignements
+        text_path (str): path to text
+        audio_path (str): path to audio
+        words (List[dict]): all alignements
```
roedoejet, line 46: alignments*

readalongs/align.py
```diff
@@ -956,54 +968,61 @@ def get_ancestor_sent_el(word_el: etree.ElementTree) -> Union[None, etree.Elemen
     return word_el
 
 
-def get_words_and_sentences(results) -> Tuple[List[dict], List[List[dict]]]:
-    """Parse xml into word and sentence 'tier' data
+def get_textual_words_and_sentences(
+    words: List[dict], tokenized_xml: etree.ElementTree
+) -> Tuple[List[dict], List[List[dict]]]:
+    """Parse xml into word and sentence 'tier' data with full textual words
```
roedoejet, line 974: What are "textual" words? I find this a bit confusing. I see. They're the actual orthographic words, not our internal ID representation. I would maybe change this to `get_orthographic_words_and_sentences`? I don't know, our internal format is still text, but it's not orthographic (for the language).

readalongs/web_api.py
```diff
@@ -203,5 +204,174 @@ def create_grammar(xml):
...
+                        <text src="hej-verden.xml#t0b0d0p0s0w1"/>
+                        <audio src="hej-verden.mp3" clipBegin="0.78" clipEnd="1.89"/>
+                    </par>
+                </body>
+            </smil>"""
+        ),
+    )
+
+
+class ConvertResponse(BaseModel):
```
roedoejet, line 261: Hm - I see why you did this - because some responses included two files, but I think we might actually want to design this a little differently. FastAPI has a built-in FileResponse object (https://fastapi.tiangolo.com/advanced/custom-response/?h=file#fileresponse) which takes care of sending appropriate headers along with the response (which we don't fully implement here). I think we should just parameterize the request for subtitles to return either a sentence or word response. More to come!

readalongs/web_api.py
```diff
@@ -203,5 +204,174 @@ def create_grammar(xml):
...
+    )
+
+
+def slurp_file(filename):
+    """Slurp a file into one string"""
+    with open(filename, "r", encoding="utf-8") as f:
+        return f.read()
+
+
+@v1.post("/convert_alignment", response_model=ConvertResponse)
```
roedoejet, line 287:
```suggestion
@v1.post("/convert_alignment/{output_format}", response_model=ConvertResponse)
async def convert_alignment(input: ConvertRequest, output_format) -> ConvertResponse:
```

This is how you do what you were suggesting. I would also kind of prefer this I think instead of bundling it with the body of the request? I mean it's certainly what you would do in a RESTful API, but this is more RPC-ish. 

See "path parameters" https://fastapi.tiangolo.com/tutorial/path-params/
Another option (and what might make sense for specifying sentences vs words on vtt would be something like:

`/convert_alignment/vtt?tier=sentences` and in that case the `tier=` would be a "query parameter" (https://fastapi.tiangolo.com/tutorial/query-params/). Then we could create an Enumaration class for the output formats that we accept and let FastAPI handle the case where somebody requests something else. See https://fastapi.tiangolo.com/tutorial/path-params/?h=enum#working-with-python-enumerations

readalongs/web_api.py
```diff
@@ -203,5 +204,174 @@ def create_grammar(xml):
...
+                file_name="aligned.eaf",
+                file_contents=slurp_file(prefix + ".eaf"),
+            )
+
+        elif output_format == "srt":
+            save_subtitles(words, parsed_xml, prefix, "srt")
+            return ConvertResponse(
+                file_name="aligned_sentences.srt",
+                file_contents=slurp_file(prefix + "_sentences.srt"),
+                other_file_name="aligned_words.srt",
```
roedoejet, line 356: So yes, just to re-iterate, my recommendations here would be: 

1. return `FileResponse` objects instead of this custom `ConvertResponse` object
2. use path parameters for each output format and use a Python Enumeration to make use of FastAPI handling exceptions
3. use query parameters to return either sentence or word subtitles for subtitle formats (with sentences being the default as this is what the vast majority of cases will want I expect)

roedoejet (changes requested):
Hey @joanise - this is awesome! Well done on your first API with FastAPI. I like the way you did it, but I have some suggestions in comments. Maybe we can discuss further on slack?

Thanks again - solid work. Well tested, and thought-out.

readalongs/align.py
joanise: Right, that would be clearer. I wanted to distinguish from ID, and I agree "orthographic" more precisely makes that distinction.

readalongs/align.py
```diff
@@ -956,54 +968,61 @@ def get_ancestor_sent_el(word_el: etree.ElementTree) -> Union[None, etree.Elemen
...
 
     Args:
-        results([TODO type]): the results object returned by align_audio
+        words: list of words with "id", "start" and "end"
+        tokenized_xml: tokenized or g2p'd parsed XML object
 
     Returns:
-        list of words, list of sentences
+        list of words, list of sentences (as a list of lists of words)
+        The returned words are dicts containing:
```
dhdaines, line 982: Orthogonal to this PR, but it would be more robust to use a NamedTuple or data class instead of a dict here.

readalongs/align.py
dhdaines: both are fairly clumsy to my ears/eyes, would recommend something like `get_word_names_and_sentences`, and `word_name` instead of `textual_word`

joanise: Hum, I'm not sure I like "name" better. That doesn't ring "the word written out" to me.
But I do really like the idea of changing `"text"` as the field name to `"id"` in the case of word IDs, and maybe keep `"text"` for orthographic words, or `"ortho"` or maybe something else.

dhdaines: how about `get_word_texts_and_sentences`?  mainly I just think `orthographic` and `textual` are much too verbose, and `ortho` is the name of a pesticide, not an API call :)

readalongs/align.py
joanise: God yes! I really don't like our dicts used like this all over the code. But I've never had the courage to change them, because they're really all over!

readalongs/align.py
joanise: Yeah, I like `get_word_texts_and_sentences`, `word["text"]`, `word_text`

readalongs/web_api.py
joanise: Hum, I like the path parameter, but I can't figure out how to keep my documentation I had for the enum values.
With `output_name` as a member of `ConvertRequest`, as an enum or a str, I can do:
```
output_format: FormatName = Field(
    example=FormatName.textgrid,
    title="Format to convert to, one of textgrid (Praat TextGrid), eaf (ELAN EAF), srt (SRT subtitles), or vtt (VTT subtitles).",
)
```

Is there a corresponding place I can put that `title` info when I do a path parameter? I tried giving the path param a default value as a Field with such a title, but it didn't put that title in the documentation. 

I guess there's always the docstring for the endpoint itself, that would do too.

Most satisfying, though, would be if my `class FormatName(Enum)` could itself have its own `title` or similar documentation...

readalongs/web_api.py
joanise: Well, it turns out I cannot use a FileResponse, because that asynchronously reads the file to send it, but I wipe the file out as the endpoint end when I clean up the tempdir, as we promise in our data privacy policy, so I'll use a Response instead like this:
```
            return Response(
                slurp_file(prefix + ".TextGrid"),
                media_type="text/plain",
                headers={"Content-Disposition": "filename=aligned.TextGrid"},
            )
```

joanise: I've also tested with `headers={"Content-Disposition": "attachment; filename=aligned.TextGrid"},` and it returns a download link with the contents, downloading to that filename, but I don't know if that's better for our use cases or not.

dhdaines: Alternately if you don't want to read the whole file (but the files are small enough that I think this is a non-issue) you can do this:  https://fastapi.tiangolo.com/advanced/custom-response/#using-streamingresponse-with-file-like-objects

joanise: Yeah, I was looking at StreamingResponse too, but I have to read it all into memory before I delete off the disk, so again there is not benefit in trying to do anything async optimized. The files are quite small, so I think whole-file fully sync is file.

readalongs/align.py
joanise: I just realized another nice thing about using just `text` is that it makes it consistent with the lxml.etree object these comes originally come from, since the text is (mostly) in the `.text` field of the etree object.

readalongs/align.py
dhdaines: this seems like a better idea (it is also what pocketsphinx and soundswallower do internally, except that of course their "text" is our "id")

joanise:
I've done all the changes we've discussed, I think this PR can be ready to merge. @dhdaines @roedoejet do you want to take a look again or can I just merge?

roedoejet:
> I've done all the changes we've discussed, I think this PR can be ready to merge. @dhdaines @roedoejet do you want to take a look again or can I just merge?

I'll have a quick look

readalongs/web_api.py
```diff
@@ -21,18 +21,23 @@
...
 import io
 import os
+from enum import Enum
+from tempfile import TemporaryDirectory
+from textwrap import dedent
 from typing import Dict, List, Optional, Union
 
 from fastapi import Body, FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import Response
```
roedoejet, line 31: why not `FileResponse`?

roedoejet (commented):
Looks good - but why use `Response` instead of `FileResponse`?

readalongs/web_api.py
dhdaines: I think it's because the file has to be deleted immediately after sending it, and there isn't a way (or at least not an obvious way) to do that with FileResponse.

readalongs/web_api.py
```diff
@@ -89,21 +93,20 @@ class AssembleResponse(BaseModel):
...
+    Returns:
+        langs as dict with language codes as keys and the full language name as
+        values, e.g.:
+        `{
+            "alq", "Algonquin",
+            "atj": "Atikamekw",
+            "lc3", "Third Language Name",
+            ...
+        }`
+    """
```
dhdaines, line 109: Ideally this, and all of the removal of encoding parameters, would be a separate PR, but git doesn't make that very easy to do in my experience...

readalongs/web_api.py
roedoejet: It should be the same no? The response (whether `Response` or `FileResponse`) should return the file created in the temporary directory. Once it does that, the file gets deleted.

dhdaines (approved):
looks good to me...

readalongs/web_api.py
dhdaines: The problem is that `FileResponse` sends it asynchronously, so the directory will get deleted (from exiting the with block) before the event loop gets around to sending it.

roedoejet: I see - in that case I guess we could just write a background task to delete the file after the response is sent https://fastapi.tiangolo.com/tutorial/background-tasks/

joanise: I first tried `FileResponse`, but it's broken for our use case: it asynchronously sends the file *after* the endpoint function exists, but I delete the temporary folder so the file is gone by the time `FileResponse` tries to read it and I just get a file-not-found error. David suggested `StreamingResponse` as an alternative, but it suffers from the same problem.

Because our data privacy policy promises to delete the file right away, I have to slurp the file into memory and send that. I dug around to find a "FileResponse from memory" solution and what I found was to put it in a plain `Response` with the headers I set.

readalongs/web_api.py
joanise: Actually, what makes it hard is that my changes are kinda mixed together. If I had decided to separate the two tasks, I could have easily done it, and I guess I should have. Git actually helps doing that, but I was just too lazy to do it. Full disclosure: I consider myself a Git wizard and there's not much I can't do with it. I think it would have been easy, but a "normal" person would probably disagree. :)

readalongs/align.py
joanise: That, if and when we do it, will definitely be in a separate PR.

joanise: But I like the idea enough that I think I'll do it.

readalongs/web_api.py
joanise: Interesting. I was worried about creating a race condition, but FileResponse accepts a `background` parameter provide a background task to run after running. See https://stackoverflow.com/a/68579536
This looks like a very elegant solution, I'll give it a shot.

joanise: Dang I hate Windows! I have it all working with a FileResponse now, but if I run the tests on Windows, the SRT, VTT and EAF files get CRLF newlines instead of LF. I guess that's valid too, but what a nuisance! It means my unit testing has to convert before testing for equality. Bleh.

My slurp_file solution was reading the file back on windows, converting it to LF in the process, but a FileResponse sends the file contents verbatim.

readalongs/web_api.py
```diff
@@ -386,6 +423,13 @@ async def convert_alignment(
                 detail="If this happens, FastAPI Enum validation didn't work so this is a bug!",
             )
 
+    except Exception:
```
roedoejet, line 419: nice

readalongs/web_api.py
```diff
@@ -322,60 +326,93 @@ async def convert_alignment(
...
+                    detail="XML+SMIL file pair provided cannot be converted",
+                ) from e
+            return FileResponse(
+                prefix + ".TextGrid",
+                background=cleanup,
                 media_type="text/plain",
-                # With "attachment;", the result has to be downloaded
-                # headers={"Content-Disposition": "attachment; filename=aligned.TextGrid"},
-                # Without "attachment;", the result is right in the response body
                 headers={"Content-Disposition": "filename=aligned.TextGrid"},
```
roedoejet, line 347: I think setting `filename="aligned.TextGrid"` is preferred and the `Content-Disposition` header will be set automatically. but this is a very minor point.

PR #120, opened 2022-10-13
Dev.pre release

joanise:
tweaks due to latest g2p release, needed before release Studio too


joanise:
Since this still works with g2p v0.5.20210825 I didn't bump the g2p version requirements

PR #124, opened 2022-10-25
fix: ignore BOMs when reading files, though never generate them

joanise:


readalongs/cli.py
```diff
@@ -625,7 +625,7 @@ def tokenize(**kwargs):
     context_settings=CONTEXT_SETTINGS,
     short_help="Apply g2p to a tokenized file, in preparation for alignment.",
 )
-@click.argument("tokfile", type=click.File("rb", encoding="utf8", lazy=True))
```
dhdaines, line 628: Hi! Why was this encoding="utf8" removed?  Doesn't seem related to the ignoring-BOMs change.

joanise: Good question. It was just tidying up. This file is in mode "rb", binary, so I believe its encoding parameter was always just ignored.

dhdaines: Ah! I didn't notice that - yes, encoding is meaningless in that case...

joanise: Unit testing already covered it, but I just tested manually and with this encoding parameter removed `readalongs g2p` continues to read utf-8 input correctly.

Working on this patch, I discovered that `etree.parse()` extensively analyzes the first line of its input and decodes accordingly. You can throw just about anything at it, binary or text, bom or not, specified encoding or not, and it will figure it out.

Beyond that, according to https://stackoverflow.com/a/21698118 all of the above is due the XML standard itself stating that the first line of an XML file must contain enough information to support the decoding, so I suspect `"rb"` is the best file mode to use when loading XML files. (The answer uses `'r'` but it's for Python 2.7.)

joanise: I also notice we have `load_xml()` functions defined in two files in `waveform2svg/` that open the file first and then use `etree.fromstring(fin.read())`, but they could be streamlined with `etree.parse(filename).get_root()` since etree.parse handles the opening of the file itself.

joanise: And... I'm going on at too much length here, but I just tested and this will *not* work on Windows:
```
with open(filename, "r") as f:
    xml = etree.parse(f).getroot()
```
Both `"rb"` and `"r", encoding="utf-8"` work fine, but throwing a misleading encoding at it makes it fail. (My default on Windows is something like cp1252, so I get a UnicodeDecodeError on the first non-ASCII character.)

deltork (approved):
- [✓]  passed tests
- [✓] able to read config file with BOM

PR #126, opened 2022-10-31
fix: ignore whitespace on blank lines for paragraph and page breaks

joanise:
Fixes https://github.com/ReadAlongs/Studio-Web/issues/14


joanise:
In the end, this PR is mostly refactoring, with three lines of code to actually do the real work. The refactoring helped me figure out what change I had to do, and makes the code better imho. I separated the commits into `refactor:` vs `fix:` for clarity and to make it possible to see just the changes required for the fix.

PR #128, opened 2022-11-08
Updates for SoundSwallower 0.4.0

dhdaines:
Yes, it is not actually released yet, but I want to make sure everything works right away!


dhdaines:
Seems to have a small problem on windows :( I'll maybe have to do a bugfix release for that

joanise:
I love it that our CI tests have caught the problems on Windows. The power of CI in action! I'm glad I added that test-on-windows suite.

readalongs/align.py
```diff
@@ -637,19 +637,14 @@ def align_audio(
...
-            try:
-                # Process raw segmentation, adjusting alignments for DNA
-                aligned_words = process_segmentation(
-                    segmentation=segmentation,
-                    curr_removed_segments=curr_removed_segments,
-                    noisewords=noisewords,
-                    frame_size=frame_size,
-                    debug_aligner=debug_aligner,
-                )
-            except RuntimeError:
```
joanise, line 649: Ah, it's nice to see you fixed this in ss.

joanise (commented):
If I understand things correctly, this PR here adjusts Studio for the changes in the SS API, without changing any semantics. You've improved the SS API and that simplified the code here, basically.

readalongs/align.py
```diff
@@ -236,30 +236,30 @@ def create_asr_config(
 
     Returns:
         soundswallower.Config: Basic configuration."""
-    asr_config = soundswallower.Decoder.default_config()
+    asr_config = soundswallower.Config()
```
joanise, line 239: I like this interface, it's much nicer.

readalongs/align.py
```diff
@@ -426,47 +423,50 @@ def align_sequence(
...
+        if word_seg.text in noisewords:
             continue
-        start = word_seg.start_frame * frame_size
-        end = (word_seg.end_frame + 1) * frame_size
-        # change to ms
-        start_ms = int(start * 1000)
-        end_ms = int(end * 1000)
+        start = word_seg.start
+        end = word_seg.start + word_seg.duration
+        # round to milliseconds to avoid imprecisions
```
joanise, line 452: Good idea.

readalongs/align.py
```diff
@@ -426,47 +423,50 @@ def align_sequence(
...
     for word_seg in segmentation:
-        if word_seg.word in noisewords:
+        if word_seg.text in noisewords:
             continue
-        start = word_seg.start_frame * frame_size
-        end = (word_seg.end_frame + 1) * frame_size
-        # change to ms
-        start_ms = int(start * 1000)
-        end_ms = int(end * 1000)
+        start = word_seg.start
```
joanise, line 450: This, and the changes to SoundSwallowerDecoderStub, implies that you're now providing start and duration in seconds (as a float) instead of frames. Definitely a nicer interface. I assume you've made sure the code is all consistent with this change, and I suppose that's unit change that unit testing is exercising extensively (I think).

readalongs/align.py
```diff
@@ -426,47 +423,50 @@ def align_sequence(
...
             start_ms += calculate_adjustment(start_ms, curr_removed_segments)
             end_ms += calculate_adjustment(end_ms, curr_removed_segments)
             start_ms, end_ms = correct_adjustments(
                 start_ms, end_ms, curr_removed_segments
             )
-            # change back to seconds to write to smil
-            start = start_ms / 1000
-            end = end_ms / 1000
-        aligned_words.append({"id": word_seg.word, "start": start, "end": end})
+        # change back to seconds
```
joanise, line 462: Argh, this back and forth between seconds and ms is confusing! But that's not your doing, it was already like that. At least the variables that hold ms are called `_ms` so we can make sense of it.

readalongs/align.py
dhdaines: it was necessary, actually :) otherwise we ended up with overlapping boundaries...

readalongs/align.py
dhdaines: yes, the important thing is to ensure that everything is rounded to milliseconds consistently...

readalongs/align.py
dhdaines: As far as I can tell everything is consistent - that's also why the class name changed from Segment to Seg, to make sure things would fail right away

readalongs/align.py
dhdaines: Yes - in JavaScript actually there is no more Config class, because configuration and initialization of the decoder are done separately.  It isn't ideal that the APIs have diverged in this way but they diverge in other ways too, largely due to the asynchronous nature of the JS code.

readalongs/align.py
joanise: Yeah, that's a good idea. Force it to break when it's semantically different, so old code can't linger around.

joanise (approved):
Looks good to me. As soon as CI passes, we can merge this, as far as I'm concerned.

joanise:
Although, are you still able to make cp37-win and cp38-win binary download files for SS? I see only cp39-win and cp310-win, and I'm pretty sure that's going to cause me headaches at some point.

dhdaines:
Oh, yes, that's no problem, I'm not sure I was making them before, but I can just add them to the list of targets for cibuildwheel.

joanise:
In fact, in the GH Actions logs, we can see that installing dependencies is a lot slower than it used to be, and we can see that it's building the wheel for SS, because our CI uses Python 3.7, so if nothing else it would make our CI recipe more efficient to have cp37-win binaries.

joanise:
You used to make more versions than now, including 37 at some point. If it's easy to do, like just a matrix setting, the full 3.7, 3.8, 3.9, 3.10, 3.11 series would certainly be useful, since these are all the currently supported versions of Python.

dhdaines:
Yes, perhaps - supporting the whole matrix of Python versions and platforms strikes me as a bit wasteful, which is why I only build 3.6, 3.8 and 3.10 on Linux as these are the versions found in Ubuntu LTS, the reasoning being if you use something more exotic, you can surely build things yourself.  But on Windows that is not always an option...  3.9 is a sensible default because that's what Anaconda is.

PR #133, opened 2022-11-14
ci: submit PR instead of pushing version bump

### LINKED ISSUE #83: CI/CD failed
It looks like the release is being created but the version bump is not updating to the master branch. See https://github.com/ReadAlongs/Studio/runs/3885903392

joanise:
Fixes #83


roedoejet (approved):
nice easy solution. Thanks @joanise

PR #134, opened 2022-11-21
ci: fix deprecation warnings for actions/create-release and set-output

joanise:
1) Replace deprecated actions/create-release by ncipollo/release-action

   https://github.com/actions/create-release is deprecated so we should
   stop using it. ncipollo/release-action is one of its maintained
   replacements, and the one our mathieudutour/github-tag-action actions
   uses in their documentation.

   Linted with https://rhysd.github.io/actionlint/ but we can't really test
   this without creating a new release, so we'll have to test it when it's
   time to create the next release.

2) ci: replace deprecated set-output and bump github-tag-action

   Refs for set-output:
    - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
    - https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter
   
   For mathieudutour/github-tag-action, we were using v5.5 which used set-output and Node.js 12, both deprecated. @v6.1 fixes both.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -28,20 +28,18 @@ jobs:
       - name: Determine tag
         id: determine_tag
         run: |
-          echo "::set-output name=TAG_VERSION::$(ls dist/readalongs-*.tar.gz | sed -e 's/.*readalongs-//' -e 's/.tar.gz.*//')"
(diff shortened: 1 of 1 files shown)


roedoejet:
since this has to do with our release, and it's fairly important this happens smoothly (ie doesn't lose our conventional commit  changes or anything like that) I unfortunately feel like we should test these changes in a test repo. Thoughts @joanise ?

joanise:
@roedoejet Just tested, and safe to merge (after ICLDC).

PR #147, opened 2023-02-06
Introduce better CORS environment variables

### LINKED ISSUE #146: Web API PRODUCTION environment variable is unnecessarily obtuse
The way of telling the web API to go into development mode makes no sense:

```
PRODUCTION= uvicorn readalongs.web_api:web_api_app --reload
```

If production mode is the default then we should instead have a `DEVELOPMENT` environment variable that enables permissive CORS headers.

The `PRODUCTION` variable means nothing to uvicorn and it means nothing to FastAPI, we are the only ones using it.  Unless there is some dark Heroku magic involved?  But even then, why would it be production mode if (a) `PRODUCTION` is unset, *or* if (b) `PRODUCTION` is set to something truthy?

dhdaines:
instead of the confusing PRODUCTION= you now do DEVELOPMENT=1 (or anything else truthy), and you can also explicitly specify the deployed origin URL with ORIGIN if you like for production mode (fixes #146)

### DIFF
--- a/readalongs/web_api.py
+++ b/readalongs/web_api.py
@@ -56,6 +56,9 @@ from readalongs.util import get_langs
 web_api_app = FastAPI()
 middleware_args: Dict[str, Union[str, List[str]]]
 if os.getenv("DEVELOPMENT", False):
+    LOGGER.info(
+        "Running in development mode, will allow requests from http://localhost:*"
+    )
     # Allow requests from localhost dev servers
     middleware_args = dict(
         allow_origin_regex="http://localhost(:.*)?",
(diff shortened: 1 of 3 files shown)


joanise:
The documentation at the top of `web_api.py` should be updated to reflect these changes.

joanise:
We'll also have to update web-component documentation that refers back to this.

dhdaines:
> The documentation at the top of `web_api.py` should be updated to reflect these changes.

Done!

joanise (approved):
Looks good. This worked for me when I spelled DEVLOPMENT correctly. 😆

readalongs/web_api.py
```diff
@@ -47,25 +54,31 @@
 
 # Create the app
 web_api_app = FastAPI()
+middleware_args: Dict[str, Union[str, List[str]]]
+if os.getenv("DEVELOPMENT", False):
+    LOGGER.info(
```
joanise, line 59: Nice, I was thinking we should add something like that!

PR #152, opened 2023-02-16
Switch from .ras to .readalong as file extension

dhdaines:


joanise (approved):
Wow, it's just insane how many places there are where we've hard-coded that file extension!

But looks good.

Issue #155, opened 2023-02-16
unpkg can be slow resolving semantic versions


joanise:
Sometimes (today, really a lot!) unpkg needs a long time to resolve `^0.2.1` whereas `0.2.1` loads nearly instantly.

Question, should we consider using fixed versions instead of `^0.2.1`, or is this just par for the course and it's life?

This problem was not very visible with `@roedoejet/readalong@^0.1.6`, but it's suddenly very visible with `@readalongs/web-component@^0.2.1`. Is there a difference between those two accounts on unpkg?

PR #156, opened 2023-02-16
handle unpkg timeouts

joanise:
Gracefully handle the situation where unpkg is slow and times out


joanise:
Related to #155 but not a fix, just adding fault tolerance to the system.

roedoejet (approved):
Looks good! This is reasonable to do in the CLI

Issue #157, opened 2023-02-16
update bundle.js and bundle.css


joanise:
With all the changes we're making, `readalongs/text/bundle.js` and `readalongs/text/bundle.css` and most likely no longer compatible with current output, so we should update them.

joanise:
It would be worth having a process to update these more easily. I hacked some code in [dev.update-bundles](https://github.com/ReadAlongs/Studio/tree/dev.update-bundles) to do it in #158 but it was still pretty much semi manual.

roedoejet:
agreed. this is done, but we'll need to do it again once we patch with https://github.com/ReadAlongs/Web-Component/pull/177

joanise:
long done, and documented in the Studio-Web README's Build&Publish section.

PR #158, opened 2023-02-16
Update bundle.css and bundle.js to @readalongs/web-component@0.2.1

joanise:
Related to #157 but that issue should remain open: when we publish the version of Web-C that will be active at ICLDC, we'll want to update the bundles again.


roedoejet (approved):
I didn't check these are actually 0.2.1 but looks good!

PR #160, opened 2023-02-24
Output the readalong file version, allow it in the DTD, and set the version to 1.0

joanise:


dhdaines:
This looks okay - though if we want to change the DTD in a truly incompatible way, it would be best to change the top-level tag's name, which is what TEI did (from TEI to TEI.2).

dhdaines (approved):

PR #161, opened 2023-02-24
Dev.publish

roedoejet:


.travis.yml
```diff
@@ -25,10 +25,6 @@ install:
   - pip3 install -r requirements.dev.txt
   - pip3 install coverage
   - pip3 install codecov
-  - pip3 install gunicorn
```
dhdaines, line 28: ah! good catch!

requirements.min.txt
```diff
@@ -15,4 +12,3 @@ pympi-ling==1.69
 python-slugify==5.0.0
 soundswallower~=0.4.1
 webvtt-py==0.4.2
-werkzeug==2.0.3
```
dhdaines, line 18: hooray :)

dhdaines (approved):
Great! This should help our codecov stats as well...

requirements.min.txt
joanise: Don't celebrate too fast, that dependency is still in g2p...

.travis.yml
joanise: Hey, why do we still have `.travis.yml` at all? We don't use Travis anymore...

PR #166, opened 2023-02-27
fix: make capture_logs work correctly with Python >= 3.9

### LINKED ISSUE #162: capture_logs() does not disable stderr logging with Python >= 3.9
test_misc.py fails because capture_logs is not actually deactivating stderr logging, as it is intended to.

Works with Python <= 3.8, just  not >= 3.9

joanise:
Fixes #162


roedoejet (approved):
Looks good, thanks @joanise - please just also remove the runtime.txt file

PR #169, opened 2023-02-28
Maximally fix dockerfile

dhdaines:
Won't be a separate PR for the maximal fix.  It's now 621MB down from 1.18GB :-)


joanise (approved):
works

PR #174, opened 2023-03-13
minor repo cleanup work

joanise:


joanise:
@roedoejet This PR is mostly trivial, the only thing to review is whether we want a "deploy to Heroku" shield with that link the live API v1 server.

PR #177, opened 2023-04-17
refactor: move get_langs() out to g2p where it really belongs

joanise:
`get_langs()` as implement in Studio is needed for both Studio and EveryVoice, and it's tightly tied to g2p itself.

 - Refactor it to g2p via https://github.com/roedoejet/g2p/pull/248
 - Make Studio use the implementation from g2p
 - But temporarily keep the Studio version in `deprecated.py` until we publish the next release of g2p, to avoid breaking code everywhere.


readalongs/web_api.py
```diff
@@ -136,13 +136,9 @@ async def langs() -> List[SupportedLanguage]:
...
     langs, lang_names = LANGS
-    return sorted(
-        (
-            SupportedLanguage(code=code, names=dict(_=lang_names[code]))
-            for code in langs
-        ),
-        key=lambda lang: lang.code,
-    )
+    return [
+        SupportedLanguage(code=code, names=dict(_=lang_names[code])) for code in langs
```
roedoejet, line 140: I guess they are sorted elsewhere now?

roedoejet (approved):
Just a small clarification question about the sorting. Looks good!

readalongs/web_api.py
joanise: They have been sorted for quite a while, where the list is generated, i.e., in get_langs(), so it's a minor optimization to not sort again.

I've added a unit test to confirm we don't undo that at some point.

joanise: And in fact, it Studio-Web, the first think we do with this list is sort it again by names instead of code, so it's silly to sort on twice on the wrong key before sending it over!

PR #178, opened 2023-05-11
bump g2p dep and adjust make_g2p accordingly

joanise:


joanise:
Let's not merge this half an hour before a live demo! OK to review now, but please don't merge right away.

joanise:
This should be safe to merge now.

PR #179, opened 2023-05-17
Use the lexicon-based eng g2p from the g2p module

joanise:
- "eng" is now (almost) g2p'd like any other language
 - we can strip all the lexicon_g2p code from Studio, as well as the cmudict copy it had
 - test suites are adjusted accordingly


readalongs/text/convert_xml.py
```diff
@@ -100,53 +97,37 @@ def convert_word(word: str, lang: str):
...
+            ) from e
+        except NoPath as e:
+            raise ValueError(
+                f'Could not g2p "{word}": no path from "{lang}" to "{output_orthography}".'
+                f'\nRun "readalongs langs" to list languages supported by ReadAlongs Studio.'
+            ) from e
+        tg = converter(word)
+        text = tg.output_string
+        if not text:
+            # TODO Convert this to a check of the kind "is the mapping a lexicon-based one".
```
dhdaines, line 115: Hmm.  This should be a method on the `Transducer`  base class in g2p, but you can do it like this:

```python
if "lexicon" in (t.mapping.kwargs.get("type") for t in converter.transducers):
```

joanise: Easy enough, thanks for the hint!

joanise: Done. I love how your refactoring in your dev.alignments PR, making Transducer and CompositeTransducer compatible, makes this code so much simpler, with `coverter.transducers` just doing the right thing in each case.

joanise: Hey, I just had a thought: an empty g2p output from any language, lexicon or rule based, is really an error situation that will cause soundswallower to fail. I could simply that code a fair bit to just
```
        if not text:
            nonlocal g2p_empty_warning_count
            if g2p_empty_warning_count < 2 or verbose_warnings:
                g2p_empty_warning_count += 1
                LOGGER.warning(
                    f'The output of the g2p process for "{word}" with lang "{lang}" is empty.'
                )
        valid = converter.check(tg, shallow=True) and text
```

joanise: The useful side effect here is that if a language's rules turn a token to nothing, it'll be tried again with und and thus we'll be much more likely to get a successful alignment.

PR #180, opened 2023-06-16
fix(deps): adjust deps for latest g2p update

joanise:
required once we merge  https://github.com/roedoejet/g2p/pull/259


roedoejet (approved):
Looks good to me. Not sure why we need click>=8.0.4 but it doesn't appear to cause any issues.

joanise:
> Not sure why we need click>=8.0.4 but it doesn't appear to cause any issues.

I'm basically aligning it with what g2p declares. What version we actually get will depend on the Python version.

PR #181, opened 2023-06-29
build: force the latest g2p@main for heroku

joanise:
The idea of this commit is that studio as published on PyPI still depends on a g2p release as declared in `requirements.min.txt`, but the live system on Heroku uses the current main branch from GitHub instead.


roedoejet (approved):
Nice! I like that we can do this just for Heroku because the requirements files are separated.

PR #182, opened 2023-07-06
feat: publish the accuracy script we used in our 2022 papers

joanise:
@littell is this all Xinjian requires?


littell (approved):
Yes, this should be enough

PR #184, opened 2023-07-25
Update search for noise dictionary to match all SoundSwallower versions (fixes #183)

### LINKED ISSUE #183: Strange IndexError caused by [SMACK]
I can't provide steps to re-create this because the data is private, but while aligning a specific audio file, an IndexError is thrown when trying to get the word element:

```
File "/Users/pinea/.pyenv/versions/3.9.14/lib/python3.9/site-packages/readalongs/align.py", line 1006, in get_word_texts_and_sentences
    word_el = get_word_element(tokenized_xml, word["id"])
  File "/Users/pinea/.pyenv/versions/3.9.14/lib/python3.9/site-packages/readalongs/align.py", line 972, in get_word_element
    return xml.xpath(f'//w[@id="{el_id}"]')[0]
IndexError: list index out of range
```

A breakpoint showed that the word in question was `{'id': '[SMACK]', 'start': 74.19, 'end': 75.66}` - removing it fixes the issue, but I presume this is something coming from SoundSwallower. @dhdaines - how should we be handling this?

dhdaines:
A healthy dose of silencing for mypy (because SoundSwallower's type annotations may be slightly bogus)

### DIFF
--- a/readalongs/align.py
+++ b/readalongs/align.py
@@ -277,7 +277,7 @@ def create_asr_config(
     # Set the minimum FFT size (no longer necessary since
     # SoundSwallower 0.2, but we keep this here for compatibility with
     # old versions in case we need to debug things)
-    frame_points = int(asr_config["samprate"] * asr_config["wlen"])
+    frame_points = int(asr_config["samprate"] * asr_config["wlen"])  # type: ignore
     fft_size = 1
     while fft_size < frame_points:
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
Speedy quick work - thanks @dhdaines !! I just tried this and it fixed the issue. I'll merge now.

PR #187, opened 2023-09-11
fix: g2p main is about to be 2.0, but we still want 1.1 on Heroku

### LINKED ISSUE #186: Revert to using only published versions of g2p
As discussed in https://github.com/roedoejet/g2p/pull/282 - there are breaking changes coming to g2p and so we should revert the commit that pulls the latest version of g2p's main branch. Eventually we should update Studio to work with g2p 2.0

joanise:
We had changed things so Heroku builds always took origin/main for g2p, but now that we're working on g2p 2.0 with breaking changes, we have to revert this change and get the latest g2p 1.1.* release instead.

Fixes #186

### DIFF
--- a/requirements.api.txt
+++ b/requirements.api.txt
@@ -3,5 +3,3 @@
 gunicorn
 # uvicorn works on all platforms and is required for both dev and prod deployments
 uvicorn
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
thanks @joanise !

PR #191, opened 2023-10-01
Actually switch to using latest g2p

dhdaines:
Not sure if there is a good reason we aren't doing this, perhaps I missed it in a comment thread somewhere.

Basically the requirements here, for whatever reason, actually result in uninstalling the latest g2p when you install Studio, which is a problem since we are now fixing things in that version of g2p...

### DIFF
--- a/requirements.min.txt
+++ b/requirements.min.txt
@@ -3,8 +3,7 @@ anyio<4.0.0
 chevron==0.14.0
 click>=8.0.4
 coloredlogs==10.0
-fastapi>=0.78.0
-# Note: when we move g2p to >=2, we will need fastapi>=0.100.0 which requires httpx
+fastapi>=0.100.0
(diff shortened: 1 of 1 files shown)


dhdaines:
Hmm, this might not be strictly necessary since one simply has to `pip install --upgrade fastapi` and things will be okay.  Not sure why the latest g2p doesn't get installed by default though.

dhdaines:
Ah, never mind, it's because I had a pre-existing installation of `pydantic` in my virtualenv and so it was not upgraded.  Closing this ... though it would be nice if we could update g2p on Heroku ...

joanise:
Right, we had studio no longer take g2p@main because for a while that was not compatible, but I believe we're compatible again, so I think we can change it back to g2p@main.

joanise:
See #192 (and review please!)

PR #193, opened 2024-01-05
Miscelaneous changes

joanise:
- optimize the CLI and report its speed in PR comments
 - use the latest setup-ffmpeg action to get the latest ffmpeg in CI
 - minor style changes suggested by ruff, that black doesn't revert


joanise:
Not sure where that `pkg_resources` comes from in the import analysis in CI, but I don't see that on my machine or on Linux. It's faster on Linux for me, at 0.13s. It was already OK before this PR, but I like this better. On Windows, it's slower: now at 0.9s, previously at 1.3s. I don't know why, but I would not be surprised if it had to do with NRC's security stack, and I expect there's little I can do about it. Still, 1s is fine, I don't mind it.

dhdaines:
Windows is just inherently slow at opening files and likely NRC's security stack (which probably has some hideous "antivirus" that runs on every. single. call. to. fopen.) isn't helping there.  Guessing that pkg_resources is the acoustic model files that we include with readalongs.

dhdaines (approved):
All looks good to me, presume that it works since tests pass!

joanise:
> Windows is just inherently slow at opening files and likely NRC's security stack (which probably has some hideous "antivirus" that runs on every. single. call. to. fopen.) isn't helping there.

Ouch, yes! Just tested on my personal machine. It's more recent by 3 years (I just bought it), but still, it runs these `-h` commands five times faster! Our security stack sucks. It's tempting to move my OSS work to my personal machine...

> Guessing that pkg_resources is the acoustic model files that we include with readalongs.

Don't think so, that would appear in every scenario, and it's a top-level import from CI. Google tells me it's part of setuptools, maybe some things are installed differently on the CI image?

PR #195, opened 2024-01-19
lock numpy < 2

joanise:
See: https://pythonspeed.com/articles/numpy-2

When numpy 2.0.0 is actually out, we'll want to test g2p and our other projects against it before allowing it.

Also fix the bold bug with coloredlogs, long ago fixed in g2p, which prevented us from updating it.


roedoejet (approved):
ty @joanise !

PR #197, opened 2024-01-31
Bump lxml to a version compatible with Python 3.11 on Windows

joanise:
This makes RAS properly compatible with Python 3.11 everywhere, although installation on Windows will require a compiler until the SoundSwallower wheels are published for 3.11 (see https://github.com/ReadAlongs/SoundSwallower/issues/60)


roedoejet (approved):
looks good, thanks

PR #204, opened 2024-02-02
docs: document how our builds on Heroku are controlled

joanise:
Last summer, I documented how our Heroku deployment is controlled, but I just noticed this file, never committed, never pushed... I checked, and I believe it's still accurate.


roedoejet (approved):
I didn't verify it - but this is in-line with my memory of how this works. Thanks @joanise

PR #205, opened 2024-02-06
style: bump black to 23

joanise:
No code changes to commit, the differences between black 22 and 23 do not currently affect the code base because I committed those changes a while ago.

When we merge this, developers should reinstall black with  `pip install -r requirements.dev.txt` in their Python environments.


dhdaines (approved):
Seems fine to me!

PR #207, opened 2024-03-19
docs: use the CL anthology link for the SIGUL paper

joanise:
The ACL anthology links are much nicer to share.


dhdaines (approved):
agreed!

PR #212, opened 2024-04-04
feat: cli generates readme

deltork:
copied and pasted readme generator from web-studio into the cli-studio


readalongs/text/util.py
```diff
@@ -22,6 +22,8 @@
 # too many times, so define it anyway.
 unicode = str
 
+CURRENT_WEB_APP_VERSION = "1.2.0"
```
joanise, line 25: We're actually at 1.2.1 now... it's hard to keep up!

readalongs/text/util.py
```diff
@@ -295,3 +298,68 @@ def parse_time(time_string: str) -> int:
         raise ValueError(
             f'cannot parse "{time_string}" as a valid time in h/m/s/ms'
         ) from e
+
+
+TEMPLATE_README_TXT = """
```
joanise, line 303: Let's have a comment here that says this text should be kept in sync with the same text in the Web-Component repo, with a pointer to where that is. And let's add a comment there pointing here. That way we'll know to update both when we update one.

joanise (changes requested):
Looks good, with minor details.

Re black: did you install the recommended pre-commit hooks? If so, black should get applied automatically for you when you commit. Unfortunately, it's not like node, where `npm install` installs husky automatically for us. It would be nice to have the same here!

readalongs/text/util.py
deltork: I will make it 1.2.x. This way we only update when we do minor releases and not for every patch

readalongs/text/util.py
deltork: DONE

readalongs/text/util.py
```diff
@@ -21,6 +21,8 @@
 # removed "try: unicode() except" block (was for Python 2), but this file uses unicode()
 # too many times, so define it anyway.
 unicode = str
+# todo: sync with web component major and minor releases
+CURRENT_WEB_APP_VERSION = "1.2.x"
```
joanise, line 25: Neat, I didn't realize you could do that with unpkg!

joanise (approved):
Thanks! Looks good.

PR #216, opened 2024-04-23
Fix pydantic issues

### LINKED ISSUE #215: web_api.py uses Pydantic but we don't declare the version we need, and we get deprecated warnings with Pydantic 2
We clearly rely on g2p requiring pydantic for us, but since Pydantic v1 and v2 are quite different, we should declare the version we expect in `requirements.txt`, and fix this deprecation warning I got from `web_api.py` with Pydantic 2.4.0 installed:

> C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\pydantic\fields.py:798: PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'example'). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.4/migration/

joanise:
- Declare our dependency on Pydantic. We're actually super relaxed: `web_api.py` breaks with Pydantic 1.7, but works fine from 1.8.2 to 2.x, so I made it that broad. I'm going to rely on FastAPI's exclusions to block out "bad" version, I don't feel the need to duplicate that here. E.g., `fastapi==0.110.1` wants `pydantic!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0,>=1.7.4` (!!!)
 - Pydantic 2 does not allow `Field(example="foo")`, only `Field(examples=["foo"])`, while Pydantic 1 is happy with either, so let's only use the latter to suppress the deprecation warning.
 - And the ancient flake8 use used in `.pre-commit-config.yaml` was broken on my machine, so I bumped it to the current version.

Fixes #215

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,6 +32,6 @@ repos:
         stages: [commit]
 -   repo: https://github.com/pycqa/flake8
     # do flake8 last to avoid duplicate reports
-    rev: 3.8.3
+    rev: 7.0.0
     hooks:
     -   id: flake8
--- a/readalongs/web_api.py
+++ b/readalongs/web_api.py
@@ -316,7 +316,7 @@ class ConvertRequest(BaseModel):
     """Convert Request contains the RAS-processed XML"""
 
     dur: Union[float, None] = Field(
-        example=2.01,
+        examples=[2.01],
         gt=0.0,
         title="The duration of the audio used to create the alignment, in seconds.",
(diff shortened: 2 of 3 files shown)


roedoejet (approved):
Looks great to me - thanks for the clear explanations in the PR description!

PR #217, opened 2024-04-26
fix: updated the exported readme.txt 

deltork:
Readme.txt WordPress section now contains
- default WordPress upload path 
-  image-asset-folder attribute


readalongs/text/util.py
```diff
@@ -1,368 +1,374 @@
...
-            TEMPLATE_README_TXT.format(
-                version=CURRENT_WEB_APP_VERSION,
-                text=tokenized_xml_basename,
-                audio=audio_basename,
-                theme=theme,
-                header=header,
-                subheader=subheader,
-            )
-        )
+###########################################
```
joanise, line 1: This PR changes the newline style on this file from LF to CRLF. Please revert that change, and maybe configure your editor to always use LF, and maybe even use `git config --global core.autocrlf input` unless you have other projects that need CRLF line ends.

joanise:
Oh, also, the tests don't pass, I assume you'll take the time to fix them as part of this PR?

deltork:
> Oh, also, the tests don't pass, I assume you'll take the time to fix them as part of this PR?

I added a comment to the XML which caused the test to fail. 
I will make a separate branch to work with you to accomplish that.

joanise:
Looking at the generated readme file, I thought it was hard to read because the structure was not obvious enough. I just pushed a commit that fixes that, with `===` marking level 1 headings and `---` level 2, to make it easier to quickly grasp the sections.

joanise (approved):
Looks good, thanks!

joanise:
> Looking at the generated readme file, I thought it was hard to read because the structure was not obvious enough. I just pushed a commit that fixes that, with `===` marking level 1 headings and `---` level 2, to make it easier to quickly grasp the sections.

And I suppose I should do the same in the web app generated readme as well...

PR #220, opened 2024-05-24
Bring the bundled web-component up to date

### LINKED ISSUE #219: Update the fallback bundles in Studio CLI's `readalongs/text/`
`bundle.js` and `bundle.css` were last updated more than a year ago, so if the CLI is used offline, created single-file HTML RAs will have an ancient web component.

We should also bump `JS_BUNDLE_URL` and `FONTS_BUNDLE_URL` in `make_package.py` to the current, which is 1.4.0 as I write this. We're still declaring `^1.0.0` there, which I guess resolves to the same thing, but still...

joanise:
### PR Goal? 

update the long stale web-component offline bundles, and current web-component version.

### Fixes? 

Fixes #219

### Feedback sought? 

just make sure it works OK

### Priority? 

normal

### Tests added? 

n/a

### How to test? 

run `cd test; readalongs align -o html data/ej-fra.txt  data/ej-fra.m4a -l fra delme` and see that the single-file HTML in delme has the gear menu (i.e., the latest web-component version) and `index.html` asks for `web-component@1.4.x`.

### Confidence? 

high

### Version change? 

We're long overdue for a release, not for this PR but for previous changes, with at least a minor bump.

Significant changes:
 - we dropped Python 3.7 and added support for 3.12
 - we added the readme in the generated directory
 - Most changes were really in the web-component, not here.

Even though dropping Python 3.7 is a breaking change, I'd be inclined to just bump this project to 1.1, not 2.0, because is really hasn't changed much.
But I don't mind going to 2.0 if we want to be strict with the breaking=>major bump semantics.


roedoejet (changes requested):
This looks fine, but I followed the steps you gave and got this:

image

deltork (approved):

joanise:
@roedoejet  I can't reproduce your error. My single-file HTML has this: `audio="data:audio/mp4;base64,AAAAG...` while your screen capture shows `audio="data:audio/mp4a-latm;base64,AAAAG...` could this be incompatible with a Mac? @marctessier any chance you can test this one your end?
Also, Aidan, do you get the same problem on `origin/main`? I suspect the issue is not related to this PR.

PR #222, opened 2024-06-19
misc refactorings

joanise:
### PR Goal? 

Merge in some refactoring I was doing while working on something I abandoned since.

Two useful refactoring changes:
 - parse XML in only one place, so we do it right consistently, make the code DRYer, and make it easier to remember how to do it right.
 - `web_apy.assemble()`: use the input_string directly with a split, instead of converting it to a StringIO just to call readlines() on it.

### Feedback sought? 

regular validation

### Priority? 

low

### Tests added? 

Already covered by existing tests.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/readalongs/web_api.py
+++ b/readalongs/web_api.py
@@ -28,7 +28,6 @@ http://localhost:8000/api/v1/docs
 
 """
 
-import io
 import os
 import tempfile
 from enum import Enum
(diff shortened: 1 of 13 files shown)


readalongs/web_api.py
```diff
@@ -48,6 +47,7 @@
 from readalongs.text.convert_xml import convert_xml
 from readalongs.text.make_dict import make_dict_list
 from readalongs.text.tokenize_xml import tokenize_xml
+from readalongs.text.util import parse_xml
```
roedoejet, line 50: wow this really was in a lot of places

roedoejet (approved):
lgtm!

PR #225, opened 2024-06-26
Adding meta tag to .readalong

deltork:
### PR Goal? 
To add meta data to read along file format

### Fixes? 
enhancement of format

### Feedback sought? 
meta data with generator information is being added to .readalong files

### Priority? 
Normal

### Tests added? 
Modified existing test with meta data

### How to test? 
`cd test; readalongs align -l fra data/ej-fra.txt data/ej-fra.wav meta_test`

### Confidence? 
High

### Version change? 
Minor

PR #226, opened 2024-06-26
feat: including generator meta tag in .readalong files

roedoejet:
Rebased duplicate of https://github.com/ReadAlongs/Studio/pull/225 - please either close this one or close #225 and re-request the reviews here.


deltork (approved):
ready for merge

joanise:
Just discovered an issue while discussing #229, repeating the comment here since it's relevant here:

> Oh, wait, no, but that uncovers a bug in versioning-patch: `VERSION` and `__version__` were previously meant the be tied, with `VERSION` being Major.minor and `__version__` being generated as Major.minor.patch. On the other hand, the `format_version` introduced here will need a new variable, not defined before, because we cannot assume it's always going to be the same as readalongs' Major.minor version: we might add features or make breaking changes that don't affect for the .readalong format, in which case we'll be the software minor and/or major, but we won't change the format version unless that too gets modified.

joanise:
Also, `cli-guide.md` has a reference to `read-along-1.0.dtd` which needs to get bumped to 1.1.

joanise (changes requested):
Please create a new constant, as mentioned above, for `format_version`.

joanise:
Since I broke this code with my other PR, I rebased it for you and created the new variable. @deltork take a look and make sure it's OK. I'll approve, but it's also worth a re-review by someone else. @roedoejet ?

deltork:
> Please create a new constant, as mentioned above, for `format_version`.

Thanks for doing this

> Since I broke this code with my other PR, I rebased it for you and created the new variable. @deltork take a look and make sure it's OK. I'll approve, but it's also worth a re-review by someone else. @roedoejet ?

Looks good!

joanise (approved):
My requested changes are done.

PR #229, opened 2024-07-09
Publish based on tag push instead of release PR

joanise:
### PR Goal? 

Revamp our publication process, and our versioning system.

 - Versions no longer have dates, they're just basic Major.minor.patch now, e.g., 1.1.0
 - To publish, we now edit `readalongs/_version.py` with the new version, tag it and push the tag

### Feedback sought? 

 - Sanity check
 - opinion on the changed versioning model

### Priority? 

before the next release

### Tests added? 

n/a

### How to test? 

manually tested via test variant, see https://github.com/ReadAlongs/Studio/actions/runs/9862260494

### Confidence? 

high

### Version change? 

minor bump, since the current patch number is >20M, and the next patch number will be 0. But we're during for a minor release anyway!


deltork (commented):
Everything looks good!
This will conflict with dev.ap/versioning-patch. We need to discuss how to proceed.

joanise:
> This will conflict with dev.ap/versioning-patch. We need to discuss how to proceed.

Yes, but it's a very simple conflict to resolve, since it's just removing VERSION from `readalongs/__init__.py` and keeping the right version in `readalongs/_version.py`.

Oh, wait, no, but that uncovers a bug in versioning-patch: `VERSION` and `__version__` were previously meant the be tied, with `VERISON` being Major.minor and `__version__` being generated as Major.minor.patch. On the other hand, the `format_version` used in versioning-patch will need a new variable, not defined before, because we cannot assume it's always going to be the same as readalongs' Major.minor version: we might add features or make breaking changes that don't affect for the .readalong format, in which case we'll be the software minor and/or major, but we won't change the format version unless that too gets modified.

.codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1%
```
roedoejet, line 5: oh cool! we should add this elsewhere

.github/workflows/matrix-tests.yml
```diff
@@ -14,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - uses: FedericoCarboni/setup-ffmpeg@v3
+      - uses: FedericoCarboni/setup-ffmpeg@v2
```
roedoejet, line 18: why?

readalongs/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.20240426"
+VERSION = "1.1.0"
```
roedoejet, line 1: we should add the same pep440 unittest that's in everyvoice.

roedoejet (approved):
looks great @joanise - please add the pep440 unittest from everyvoice here as well to ensure `readalongs/_version.VERSION` stays pep440 compliant

.github/workflows/matrix-tests.yml
joanise: I left a comment in the commit, but I guess that's not enough:
https://github.com/ReadAlongs/Studio/pull/229/commits/93d68daf8a217f58aa1bb505b0dfdb13266868f6
The reason is we regularly have failed CI runs that fail on downloading ffmpeg. Going back to v2 gets ffmpeg internally from github, so it'll be a lot more reliable, at the fairly minor cost of not being the latest.

.codecov.yml
joanise: Yeah, it's annoying to get a CI fail for a project drop of 0.01%

joanise: especially when it's because you just deleted a covered line but patch coverage is 100%!

PR #232, opened 2024-07-12
fix: added optional id to meta tag attributes

deltork:
### PR Goal? 
fixing read-along DTD version 1.1 . The XML package is adding id to meta tag, adjusted the DTD to allow meta to have id attributes 

### Fixes? 
fix DTD validation error

### Priority? 
HIGHEST

### Tests added? 
n/a

### Confidence? 
HIGH

### Version change? 
n/a


joanise (approved):
Great, thanks for catching this!

PR #233, opened 2024-07-15
test: added backwards compatibility test for DTD

deltork:
### PR Goal? 

Adding test case to check backwards compatibility of version 1.1 DTD

### Fixes? 
n/a

### Feedback sought? 
sanity check

### Priority? 
LOW

### Tests added? 
test case added to test_DTD

### How to test? 
`cd test; python test_dtd.py`

### Confidence? 
HIGH

### Version change? 
n/a

### DIFF
--- a/test/test_dtd.py
+++ b/test/test_dtd.py
@@ -8,6 +8,8 @@ from unittest import TestCase, main
 
 from lxml import etree
 
+from readalongs.text.util import load_xml
+
 DTDPATH = os.path.join(
     dirname(__file__), "..", "readalongs", "static", "read-along-1.1.dtd"
 )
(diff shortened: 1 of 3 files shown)


test/test_dtd.py
```diff
@@ -58,6 +58,36 @@ def test_invalid_inputs(self):
                     self.dtd.validate(parsed), f"{name} validates but shouldn't"
                 )
 
+    def test_backwards_compatibility(self):
+        # the DTD needs to be backwards compatible as long as the major version does not change
+        versions = """
+                ras-dtd-1.0.readalong
+                ras-dtd-1.1.readalong
+            """.strip().split()
```
joanise, line 66: That's a funny choice. Why not simply `versions = ["ras-dtd-1.0.readalong", "ras-dtd-1.1.readalong"]`?

test/test_dtd.py
```diff
@@ -58,6 +58,36 @@ def test_invalid_inputs(self):
...
+        for name in versions:
+            path = os.path.join(dirname(__file__), "data", name.strip())
+            # DTD is text, XML is binary... okay
+            with open(path, "rb") as infh:
+                parsed = etree.parse(infh)
```
joanise, line 71: We factored out how to correctly and safely load an XML file in `load_xml`. `test_dtd.py` should not use `etree.parse` at all, please add `from readalongs.text.util import load_xml` and use `parse_xml` everywhere we load an XML file in this test file.
`etree.DTD` is appropriate to use, but not `etree.parse`.
This will make the code shorter, too, since `load_xml()` accepts a file name and handles opening it for you.
We also have `parse_xml()` for when the XML comes from a string in memory instead of a file.

Surprisingly, loading XML files the default way can actually be a vulnerability, due to XML bombs (Google it, those are both cool and disturbing!), so it's not just for making the code DRY or shorter, our factored out functions are actually safer.

joanise (commented):
Thanks for adding this! See some change requests below, though.

test/test_dtd.py
joanise: yes please

test/test_dtd.py
```diff
@@ -58,6 +60,42 @@ def test_invalid_inputs(self):
...
+            path = os.path.join(dirname(__file__), "data", name.strip())
+            # DTD is text, XML is binary... okay
+            with open(path, "rb") as infh:
+                try:
+                    parsed = load_xml(infh)
+                    self.assertTrue(
+                        self.dtd.validate(parsed), f"{name} does not validate"
+                    )
+                except etree.ParseError as e:
+                    self.fail("Error parsing XML input file %s: %s." % (path, e))
```
joanise, line 76: That's nice and clean, making things explicit, but generally, unittest handles this for us: every uncaught exception is automatically counted as a test failure by unittest.

joanise (approved):
Approving now so you can merge when it's done, but if you can replace the other etree.parse instances before merging that would be great. Thanks for these tests!

test/test_dtd.py
deltork: I hope this will help debugging (especially in the CI).

joanise (approved):
Great, thanks!

PR #236, opened 2024-07-29
Feature: Updated DTD to support annotations

deltork:
### PR Goal? 
The goal of the PR is to provide support for annotation layers in the read along. If a .readalong has annotation layers, it should be acceptable to the CLI

### Fixes? 
n/a

### Feedback sought? 
sanity check

### Priority? 
high

### Tests added? 
yes

### How to test? 
`cd test/data; readalongs align -o html -f -l fra ej-fra-annotated.readalong ej-fra.m4a ej-fra-annotated`

### Confidence? 
High

### Version change? 
n/a

 required for https://github.com/ReadAlongs/Studio-Web/pull/330


dhdaines (changes requested):
This seems good!  Except, should we increment the version number on the DTD?

deltork:
> This seems good! Except, should we increment the version number on the DTD?

I will bump up the format version to 1.2

dhdaines (approved):
Seems good now!

joanise:
I'm guessing CI fails here due (at least in part) to the panphon issue. Releasing g2p 2.1.0 or restricting `panphon<0.21` might be needed temporarily.

joanise:
I fixed CI now. There were two problems, in the end:
 - panphon 0.21 is not compatible with Python 3.8. This problem just fixed itself automatically thanks to https://github.com/roedoejet/g2p/pull/387 being merged.
 - `test/data/cs-ref.readalong` also needed to have the DTD version bumped. I've fixed that, and also made the CI output more verbose so I don't have to hunt as hard next time this test breaks.

PR #237, opened 2024-08-19
fix(deps): exclude panphon 0.21 not compat with Python 3.8

joanise:
### PR Goal? 

Deliberately exclude panphon 0.21 since it's not compatible with Python 3.8

### DIFF
--- a/requirements.min.txt
+++ b/requirements.min.txt
@@ -4,7 +4,7 @@ chevron==0.14.0
 click>=8.0.4
 coloredlogs>=10.0
 fastapi>=0.100.0
-g2p>=1.1.20230822, <2.1
+g2p>=1.1.20230822, <3
 httpx>=0.24.1
 lxml==4.9.4
 networkx>=2.6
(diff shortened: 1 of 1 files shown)


dhdaines (approved):
Good idea to raise the max g2p version.  Are we still compatible with g2p 1.1?

It's not wonderful to have a dependency on panphon which Studio doesn't actually use directly, but if we don't want to require the upcoming version of g2p as a minimum then I suppose we have to.

joanise:
Eventually I'll want to have it as `panphon>=0.19, !=0.21.0, !=0.21.1, !=0.21.2` assuming they release a patch with Aidan's PR.

And yeah, it's unfortunate to have to declare a g2p dependency here, but oh well is how I feel about it. Although maybe we should run a test: if studio never triggers the panphon import, maybe this PR in unnecessary? But no, we use panphon to confirm g2p output is valid IPA, I'm pretty sure that gets used by Studio.

joanise:
And yes, I'm pretty sure we're still compatible with g2p 1.1.

PR #238, opened 2024-08-19
fix: more places where the current dtd 1.2 version is needed

joanise:
I believe these additional places need the version bumped to match the latest dtd version too, but I'd like confirmation I'm not misinterpreting anything here.


joanise:
These changes are further fixes to #236

dhdaines (approved):
Yes, thanks, we should definitely make sure the documentation matches the version number!

Issue #240, opened 2024-09-13
The process to upload to PyPI in pythonpublish.yml is broken


joanise:
See https://github.com/ReadAlongs/Studio/actions/runs/10854351775/job/30124835565
```
ERROR    HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/          
         Invalid or non-existent authentication information. See                
         https://pypi.org/help/#invalid-auth for more information.              
Error: Process completed with exit code 1.
```

Looks like we'll have to move from `twine upload` to the same process we use for g2p, with trusted publishing, which will let us activate sigstore signing the same way as in g2p.

Although I have a question about that: I cannot see where those .sigstore files end up. Why sign if you can see and validate the signature? @SamuelLarkin maybe you know?

Assigning this to Aidan (for activating trusted publishing) and I (for updating the pythonpublish.yml workflow).

joanise:
Fixed by #241 and AP fixing the required secrets.

PR #241, opened 2024-09-13
fix: fix (or attempt to fix) the pypi publication process

joanise:
Nothing works today. 🙁
Can't publish studio.
Finding last minutes bugs in studio-web.
Bleh.

Here, I'm guessing the way we published to pypi is no longer supported. See #240.

This code is mimicked from g2p, but not tested. Please review carefully.

Before this can work, we'll need to add the PYPI_API_TOKEN to this repo's secrets, and we can probably remove the obsolete PyPI username and password secrets since they don't work anymore anyway.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -41,10 +41,17 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install build twine
+          pip install build
 
       - name: Build
         run: python -m build --sdist --wheel
+
+      - name: Sign the dists with Sigstore
+        uses: sigstore/gh-action-sigstore-python@v3.0.0
+        with:
+          inputs: >-
+            ./dist/*.tar.gz
+            ./dist/*.whl
 
       - name: Update CHANGELOG
         id: changelog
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonpublish.yml
```diff
@@ -61,8 +68,8 @@ jobs:
...
 
-      - name: Publish to PyPI
-        env:
-          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        run: twine upload dist/*
+      - name: Publish distribution 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
joanise, line 74: Based on https://docs.pypi.org/trusted-publishers/ I bet we can (and should) remove this password field.

joanise: But it works like this in g2p, so I'll leave it this way for now.

.github/workflows/pythonpublish.yml
```diff
@@ -61,8 +68,35 @@ jobs:
...
+          verbose: true
+
+  deploy-docs:
```
joanise, line 77: follows what we do in g2p

PR #242, opened 2024-10-01
fix: catch exceptions due to failure to create XML from text

joanise:
I don't know what causes this, but we got a few 5xx errors from the Heroku logs with this trace:
```
   File "/app/readalongs/web_api.py", line 209, in assemble
     parsed = parse_xml(
              ^^^^^^^^^^
   File "/app/readalongs/text/util.py", line 138, in parse_xml
     return etree.fromstring(
            ^^^^^^^^^^^^^^^^^
 lxml.etree.XMLSyntaxError: PCDATA invalid Char value 3, line 8, column 26
```

While it makes no sense we should fail to parse XML we just generated in this broader block:
```
    parsed = parse_xml(
        create_ras_from_text(
            (request.input_text or "").splitlines(keepends=True),
            text_languages=request.text_languages,
        )
    )
```
let's just put a try/except in there since we saw it live on Heroku.

Whatever the error in the input is, though I cannot reproduce it, should generate a 422 HTTP code.

### PR Goal? 

Fix the 5xx errors we saw on Heroku today.

### Feedback sought? 

sanity check

### Priority? 

high

### Tests added? 

can't reproduce, so none

### How to test? 

beats me, I can't figure out what caused this to happen on Heroku in the first place.

### Confidence? 

high, because it's a safe patch: there's no risk in adding a try/except block that should never get triggered.

I still wish I could reproduce it, though, and thus unit test it.

### Version change? 

no

### DIFF
--- a/readalongs/web_api.py
+++ b/readalongs/web_api.py
@@ -206,12 +206,18 @@ async def assemble(
                     ),
                 )
         elif request.mime_type == InputFormat.TEXT:
-            parsed = parse_xml(
-                create_ras_from_text(
-                    (request.input_text or "").splitlines(keepends=True),
-                    text_languages=request.text_languages,
+            try:
+                parsed = parse_xml(
+                    create_ras_from_text(
+                        (request.input_text or "").splitlines(keepends=True),
+                        text_languages=request.text_languages,
+                    )
                 )
-            )
+            except etree.ParseError as e:
+                raise HTTPException(
+                    status_code=422,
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
looks good. Maybe ask the user to send us the text?

joanise:
I guess it's unlikely the user would see the actual text of the error message, but I guess it would not hurt.

PR #247, opened 2024-10-10
feat: bump RAS format to 1.3

deltork:
### PR Goal? 

To fix an issue created by the editor

### Fixes? 
https://github.com/ReadAlongs/Studio-Web/issues/347

### Feedback sought? 
sanity check

### Priority? 
High

### Tests added? 
yes

### Confidence? 
High

### Version change? 
no


readalongs/static/read-along-1.3.dtd
```diff
@@ -0,0 +1,109 @@
...
+  xml:lang CDATA #IMPLIED
+  lang CDATA #IMPLIED
+  id CDATA #IMPLIED
+  class CDATA #IMPLIED
+  do-not-align CDATA #IMPLIED
+  time CDATA #IMPLIED
+  dur CDATA #IMPLIED
+  annotation-id CDATA #IMPLIED
+  sentence-id CDATA #IMPLIED
+  xmlns CDATA #IMPLIED>
```
joanise, line 80: I'm not sure why we need xmlns here. I get it for the top-level read-along element, and I would get it if every element needed to allow it, but only one additional element I don't get.

readalongs/static/read-along-1.3.dtd
```diff
@@ -0,0 +1,109 @@
...
+
+<!ELEMENT w (#PCDATA|span|syl)*>
+<!ATTLIST w
+  xml:lang CDATA #IMPLIED
+  effective-g2p-lang CDATA #IMPLIED
+  lang CDATA #IMPLIED
+  id CDATA #IMPLIED
+  class CDATA #IMPLIED
+  do-not-align CDATA #IMPLIED
+  ARPABET CDATA #IMPLIED
```
joanise, line 90: can we make this case insensitive?

readalongs/static/read-along-1.3.dtd
```diff
@@ -0,0 +1,109 @@
...
+  dur CDATA #IMPLIED>
+
+<!ELEMENT syl (#PCDATA|span)*>
+<!ATTLIST syl
+  xml:lang CDATA #IMPLIED
+  lang CDATA #IMPLIED
+  id CDATA #IMPLIED
+  class CDATA #IMPLIED
+  do-not-align CDATA #IMPLIED
+  ARPABET CDATA #IMPLIED
```
joanise, line 101: ditto?

readalongs/text/make_package.py
```diff
@@ -20,13 +20,17 @@
 
 from lxml import etree
 
-from readalongs._version import VERSION
+from readalongs._version import CURRENT_WEB_APP_VERSION, VERSION
```
joanise, line 23: nice idea, moving `CURRENT_WEB_APP_VERSION` to `_version.py` and reusing it here.

joanise (changes requested):
Nice work, but requires a few changes.

For CI to pass on Windows (see https://github.com/ReadAlongs/Studio/actions/runs/11276460698/job/31360285739?pr=247) you also need to update `test/data/cs-ref.readalong`. I should change that test so it only looks at the line where not outputting utf-8 correctly would make a difference, instead of diffing the whole file...

More comments inline.

readalongs/static/read-along-1.3.dtd
deltork: I was thinking the same. However, I am worried about backwards compatibility. Attributes in HTML are case insensitive but are case sensitive in XML (.readalong is a subset of XML). We need a good look at the whole pipeline before we make the switch. We use both XML and HTML parsers in the various parts of the pileline. As far as I know, only the CLIs consume the ARPABET attribute.

joanise: Maybe an option is to put both `ARPABET` and `arpabet` is the dtd?

PR #251, opened 2024-11-12
Dev.ej/py313

joanise:
### PR Goal? 

Support Python 3.13

### Fixes? 

Addresses compatibility problems between some of our dependencies and Python 3.13

 - `audioop` has been removed from Python 3.13 standard libraries, but `pydub` still requires it. The community has rallied to fix this, in the form of `audioop-lts`, which we only need for Python >= 3.13
 - lxml 4.9.4 doesn't support Python 3.13. I'm not sure why we were locking the version anyway, so I just made it `>=` instead of `==`. It to work just fine. The several lxlm 5.x releases don't list any breaking changes.

### Feedback sought? 

sanity checking

### Priority? 

medium

### Tests added? 

yes, added Py 3.13 to matrix testing (and it passes)

### How to test? 

install on Python 3.13 and have fun

### Confidence? 

high

### Version change? 

maybe a patch bump so `pip install readalongs` will work for Python 3.13.

### Related PRs

https://github.com/ReadAlongs/SoundSwallower/pull/67

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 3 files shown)


roedoejet (approved):
lgtm

PR #256, opened 2024-12-17
Dev.ej/convert to offline html

joanise:
### PR Goal? 

Add an API function to convert a sequence of tokens to an Offline HTML file.

### Fixes? 

Related to https://github.com/EveryVoiceTTS/EveryVoice/issues/439

### Feedback sought? 

general review and validation

### Priority? 

EV beta

### Tests added? 

yup

### How to test? 

The easiest is to exercise via EV: see https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/103

### Confidence? 

pretty high

### Version change? 

yeah, we've been adding features, so a minor bump would make sense soon. It'll be required before publishing EV with a requirement on this feature.

### DIFF
--- a/readalongs/api.py
+++ b/readalongs/api.py
@@ -34,12 +34,17 @@ All API functions return the following tuple: (status, exception, log)
 
 Additional API function:
 
-convert_to_readalong(sentences: Sequence[Sequence[Token]], language: Sequence[str]) -> str:
+convert_prealigned_text_to_readalong():
     convert a list of sentences into a readalong XML string ready to print to file.
     Just like align and make_xml, this function expects a black line (empty list) to
     make a paragraph break, and two consecutive blank lines to make a page break.
     Unlike the other functions here, this function is not a wrapper around the CLI and
-    it just returns the string, non status.
+    it just returns the string, with no status.
+
+convert_prealigned_text_to_offline_html():
+    same as convert_prealigned_text_to_readalong, but also creates an offline HTML file.
+
+See their respective docstrings for more details.
 """
 
 import io
@@ -218,7 +223,7 @@ class Token:
         self.is_word = is_word if is_word is not None else bool(time is not None)
 
 
-def convert_to_readalong(
+def convert_prealigned_text_to_readalong(
     sentences: Sequence[Sequence[Token]],
     language: Sequence[str] = ("und",),
 ) -> str:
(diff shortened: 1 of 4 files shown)


readalongs/api.py
```diff
@@ -259,3 +265,47 @@ def convert_to_readalong(
     ).decode("utf8")
 
     return xml_text + "\n"
+
+
+def convert_to_offline_html(
```
roedoejet, line 270: suggest renaming "convert_tokens_to_offline_html" since it's ambiguous what is being converted from the fn name

test/test_api.py
```diff
@@ -152,6 +154,23 @@ def test_convert_to_readalong(self):
         readalong = re.sub(r"dur=\".*?\"", 'dur="ddd"', readalong)
         self.assertEqual(readalong, align_result)
 
+    def test_convert_to_offline_html(self):
+        html, _ = api.convert_to_offline_html(
+            self.sentences_to_convert,
+            str(self.data_dir / "noise.mp3"),
+            subheader="by Jove!",
```
roedoejet, line 161: 😆

readalongs/api.py
```diff
@@ -259,3 +265,47 @@ def convert_to_readalong(
...
+        subheader: optional subheader, will fill the readalong <span slot='read-along-subheader'>
+
+    Returns:
+        (html_contents, readalong_contents):
+         - the readalong Offline HTML file contents, ready to print to .html
+         - the readalong XML file contents, ready to print to .readalong
+    """
+
+    readalong_xml = convert_to_readalong(sentences, language)
+    try:
```
roedoejet, line 299: this is really nice and clean! well done @joanise

roedoejet (approved):
Looks good to me! A minor function renaming suggestion, but approving here anyways.

readalongs/api.py
joanise: I get the point, but I actually disagree: "convert tokens" simplifies too much and deviates more than I'm confortable with from what this function actually converts, which is "a list of sentences/paragraphs/pages of tokens". I'd rather leave it out of the function name and have the docstring and argument list clarify.

I'd be happier with `convert_prealigned_text_to_offline_html` if we want a better function name, and then I would also change the other function to `convert_prealigned_text_to_readalong`.

joanise: done

PR #259, opened 2025-01-29
refactor align helper functions out of align.py

### LINKED ISSUE #170: Refactoring idea: change web_api so it does not import soundswallower
Currently, there are three text-manipulation functions that `web_api.py` imports from `align.py`:

    from readalongs.align import create_ras_from_text, save_label_files, save_subtitles

and because of that, soundswallower is a dependency even for the web-API, even though it does not use it.

Refactoring proposal:
 - move these three functions to a more appropriate location, maybe under `readalongs/text`
 - change `requirements.txt` so it does not install soundswallower by default, making it optional, only required for using the CLI

Benefit: our Heroku image will be smaller, and any Docker image to run the web-api too.
Side node: the Docker image can also skip ffmpeg, only the CLI needs that.

joanise:
### PR Goal? 

align.py is a monster file doing too much stuff. This PR takes the alignment I/O out to align_utils.py.

The work was inspired by #170, which is non essential, but I decided to do it because it makes the code somewhat cleaner, and the Heroku image leaner.

~Not done: I'm not sure how to declare, when updating `requirements.txt` from `pyproject.toml`, to exclude soundswallower...~ (added documentation to do it on update).

### Fixes? 

Fixes #170

### Priority? 

low

### Tests added? 

This was almost all already covered by existing tests. I just had to do minor tweaks on the heroku workflow to streamline it.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,5 +146,5 @@ jobs:
 
       - name: unit test the web API
         run: |
-          pip install -e .[dev]
+          pip install httpx
           python test/test_web_api.py
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 # Heroku expects to run your app in place, but doesn't actually do this for you
-pip install -e .
+pip install -e . --no-deps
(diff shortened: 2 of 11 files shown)


roedoejet (approved):
straightforward refactor - thanks @joanise !

PR #260, opened 2025-02-03
perf: cache the bundles after we fetched them once

joanise:
### PR Goal? 

For a process using `convert_prealigned_text_to_offline_html()` in the Studio API to create a number of different readalongs, fetching the bundles is pretty slow, and has to be done repeatedly for each readalong created. Running the EveryVoice demo on our cluster, that represented an additional 3 second delay each time I clicked the synthesize button, on top of the actual synthesis time.

The solution I adopted here is simply to save the bundle contents in a file-global variable in `make_package.py`, so we only ever try to fetch each of the two bundle files once.

Memory cost: the bundles take 600KB together, so after the first fetch I'm reserving .6MB for this cache, which is pretty trivial.

### Feedback sought? 

sanity checking

### Priority? 

low

### Tests added? 

~code already covered~

Yes

### How to test? 

Run the everyvoice demo with this branch of Studio vs main, and see synthesis run faster the second and subsequent time.

### Confidence? 

high

### Version change? 

Patch, but only when we decide we really want this included in the EV demo...

### DIFF
--- a/readalongs/text/make_package.py
+++ b/readalongs/text/make_package.py
@@ -150,6 +150,10 @@ def fetch_bundle_file(url, filename, prev_status_code):
 
 _prev_js_status_code: Any = None
 _prev_fonts_status_code: Any = None
+# Cache the bundle contents so we don't fetch it more than once when running a process
+# that might generate several HTML files via the API, e.g., the EveryVoice demo app.
+js_bundle_contents = None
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
Nice optimization @joanise - lgtm

PR #261, opened 2025-02-12
Put a time limit to g2p conversion in the web api

joanise:
### PR Goal? 

Once in a while, we get Heroku errors of unknow nature. I believe they are caused by files that take too long to g2p.

We've reduced the maximum size in Studio-Web, but this PR tries to solve it with a time limit.

### Fixes? 

Attempt a fixing Heroku crashes.

### Feedback sought? 

- Review code
- The message the user sees in Studio-Web is this, with the detailed message only in English. Is that OK?
image image 
(The 5s limit was for testing, it's 45 in the PR.)
- Feedback on general approach

### Priority? 

medium

### Tests added? 

yup

### How to test? 

change `web_api.G2P_TIME_LIMIT_IN_SECONDS` to something short, like 1 or 5.

`./run-web-api.py` and navigate to http://localhost:8000/api/v1/docs and give assemble a big piece of text to process (with newlines escaped): see 422 status returned

Spin up Studio-Web locally, and paste a big piece of random text in the text box, see the toast above shown when you click on "Go to the next step!".

### Confidence? 

high that this code is good, medium that it'll fix our occasional crashes in Heroku for good.

### Version change? 

no


roedoejet:
Thanks @joanise - before I review this, won't a time-limit based exception like this penalize slower connection speeds?

joanise:
> Thanks @joanise - before I review this, won't a time-limit based exception like this penalize slower connection speeds?

No, the time limit is internal to Heroku: from starting to do g2p processing to finishing to do g2p processing.

joanise:
But slow connections raise an interesting question: what happens when the web server has a 60s timeout? Most do, there's probably nothing we can do to increase it.

I set 45s max for g2p, because the rest of the processing is much faster, just a few seconds for something that could take that long to g2p, but if sending the results starts at 55s and would take 10s, does the connection get interrupted mid-download? Or does the 60s timeout apply to starting to return a response?

joanise:
OK, 45s won't do the trick. after 24s, Heroku logged this:
```
2025-02-12T20:48:53.995361+00:00 app[web.1]: [2025-02-12 20:48:53 +0000] [2] [CRITICAL] WORKER TIMEOUT (pid:10)
2025-02-12T20:48:54.010227+00:00 app[web.1]: [2025-02-12 20:48:54 +0000] [2] [ERROR] Worker (pid:10) was sent SIGABRT!
```

So I'll have to cap g2p time around 20s! :(

joanise:
Hmmm, `gunicorn` has a default timeout of 30s, also.

https://docs.gunicorn.org/en/latest/settings.html#timeout

joanise:
@roedoejet I have just significantly augmented this PR. I'm keeping it as a single PR, because it's all about optimizing Heroku and making requests a) faster, and b) fail with a nicer error when they fail.

But now this changes:
 - cache g2p results, so we don't process the same word twice in the same document
 - give g2p a 25s budget, which is the best I can do since Heroku has a hard, non-configurable limit of 30s
 - make other fixes suggested by Heroku logs

Full details in the commit log, each commit explains what it does.

For live testing, this is temporarily deployed to 
https://ras-test-2c078d4fd6e9.herokuapp.com/api/v1/docs
with two workers (instead of our usual 5, because I was busting the memory limit on a 1x).

With all these changes, even testing with crazy big documents, I cannot generate a 503, I always get either a 200 (much more often than before) or a 422 with an informative message.

In terms of performance, I have a large document with about 2k unique types, 25k total tokens, and it takes 10s with lang=und, 20s with lang=dan, and times out after processing 21k tokens with lang=eng. This may have an unrealistically low type:token ration (because whole blocks are repeated, actually), but it still illustrates how much faster things can now be.

PR #265, opened 2025-04-07
Several improvements to the HTML files created by RAS CLI

joanise:
### PR Goal? 

This work was triggered by noticing that a .html file someone just sent us was hard to analyse by hand, in terms of what versions of things were used. So I decided to add newlines so that the text part of stuff is still visible when you disable line-wrapping, and add more versioning information to the generated files. There will be a similar PR for S-Web soonish.

While I was there, I did some clean up and refactoring and added the instructions for opening the offline HTML file from Google Drive, copied from S-Web.

### Feedback sought? 

General validation

### Priority? 

normal

### How to test? 

Use the CLI to create a readalong, with `-o html`, and look at the generated `www/index.html` and `Offline-HTML/<name>.html`.

### Confidence? 

good

### Version change? 

no

### DIFF
--- a/readalongs/text/make_package.py
+++ b/readalongs/text/make_package.py
@@ -14,6 +14,7 @@
 ###################################################
 
 import os
+import re
 from base64 import b64encode
 from mimetypes import guess_type
 from typing import Any, Union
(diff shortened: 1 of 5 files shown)


readalongs/text/make_package.py
```diff
@@ -164,28 +210,33 @@ def create_web_component_html(
     subheader=DEFAULT_SUBHEADER,
     theme="light",
 ) -> str:
-    global js_bundle_contents
+    # 6 spaces of indentation so both bundles are aligned in the BASIC_HTML template
+    def indent(text: str) -> str:
```
dhdaines, line 214: Use `textwrap.indent` instead!

joanise: Good point, thanks for the pointer!

PR #269, opened 2025-04-17
fix: minimal changes to make the Dockerfile work again

### LINKED ISSUE #257: Update Dockerfile
Our Dockerfile is out of date in several ways:
 - image is old
 - the instructions in README.md and Dockerfile don't quite align
 - `docker build .` issues eight warnings.

warnings from build:
```
8 warnings found (use docker --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 28)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 35)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 36)
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 38)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 8)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 3)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 27)
```

Update the Dockerfile, or delete it if we don't plan to maintain it.

joanise:
### PR Goal? 

Minimal changes required to make the Dockerfile work again.

### Fixes? 

Fixes #257

### Feedback sought? 

basic validation

### Priority? 

low

### Tests added? 

no, but I suppose we could create a CI workflow that builds and runs the docker image. Not sure it's worth it, though, this is really an edge use case.

### How to test? 

```
docker build . --tag=readalong-studio
docker run --rm -p 8000:8000 -it readalong-studio
```

now runs without errors or warnings.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM alpine:3.17.2 as runtime
+FROM alpine:latest AS runtime
 
-ENV APPHOME /opt/readalong-studio
+ENV APPHOME=/opt/readalong-studio
 
 # Lean, optimized installation of system dependencies
 RUN apk add python3 py3-numpy py3-yaml git ffmpeg
 
-FROM runtime as build
+FROM runtime AS build
 WORKDIR $APPHOME
 RUN apk add python3-dev py3-pip gcc g++ musl-dev ninja
 RUN python3 -m venv --system-site-packages $APPHOME/venv
@@ -13,6 +13,8 @@ RUN git clone https://github.com/roedoejet/g2p.git
 COPY requirements*.txt $APPHOME/
 RUN . $APPHOME/venv/bin/activate \
     && python3 -m pip install -r $APPHOME/requirements.txt
+RUN . $APPHOME/venv/bin/activate \
+    && python3 -m pip install soundswallower
 RUN cd $APPHOME/g2p \
     && . $APPHOME/venv/bin/activate python3 -m pip install -e .
 # Do this after all the above so we don't needlessly rebuild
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,18 +1,20 @@
...
+FROM runtime AS build
 WORKDIR $APPHOME
 RUN apk add python3-dev py3-pip gcc g++ musl-dev ninja
 RUN python3 -m venv --system-site-packages $APPHOME/venv
 RUN git clone https://github.com/roedoejet/g2p.git
 COPY requirements*.txt $APPHOME/
 RUN . $APPHOME/venv/bin/activate \
     && python3 -m pip install -r $APPHOME/requirements.txt
+RUN . $APPHOME/venv/bin/activate \
+    && python3 -m pip install soundswallower
```
dhdaines, line 17: it wasn't in the requirements.txt?

dhdaines (approved):
looks good (I avoid Docker whenever possible so I haven't actually tested it!)

Dockerfile
joanise: It's not, because heroku doesn't do alignment and therefore it does not need soundswallower. It's only in pyproject.toml.

joanise: But I want to install it early to optimize docker caching effects.

dhdaines: ah ok!

PR #271, opened 2025-05-12
build(deps): lock click<8.2.0 since it has breaking changes

joanise:
quick fix for breaking changes in click.


roedoejet (approved):
lgtm

PR #282, opened 2025-07-23
upgrade  starlett from 0.46.2 --> 0.47.2 

marctessier:
### PR Goal? 

We got a dependabot alert:

Dependency
starlette	
Version
< 0.47.2	
Upgrade to
~> 0.47.2
Defined in
requirements.txt
Vulnerabilities
CVE-2025-54121 Moderate severity

### Fixes

CVE-2025-54121

### Priority?  

We should release pretty soon.  --> Moderate severity

### Tests added? 

None

### How to test? 

I ran all tests with success

### Confidence? 

High

### Version change? 

no

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -107,7 +107,7 @@ sniffio==1.3.1
 #soundswallower==0.6.5
     # via hatch.envs.prod
 
-starlette==0.46.2
+starlette==0.47.2
     # via fastapi
 text-unidecode==1.3
     # via


joanise:
This breaks the heroku configuration, which is what `requirements.txt` is for. -- see the CI logs.

To test this, you need to install with `pip install -r requirements.txt`, and then run `bin/post_compile`,  not with the usual process.
See https://github.com/ReadAlongs/Studio/blob/main/.github/workflows/tests.yml#L141 for the process you need to reproduce locally, which itself reproduces how we install Studio on Heroku.

joanise:
From the CI logs, fastapi is a problem:
```
ERROR: Cannot install -r requirements.txt (line 48) and starlette==0.47.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested starlette==0.47.2
    fastapi 0.115.12 depends on starlette<0.47.0 and >=0.40.0
```

PR #298, opened 2026-03-30
build(deps): rebuild the Heroku requirements.txt file fresh

joanise:
### PR Goal? 

Refresh our production requirements, prompted in part by various CVEs fixed in gunicorn until version 25.3.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,8 @@
 # - requests>=2.31.0
 # - soundswallower~=0.6.0
 # - webvtt-py==0.5.1
-# - readalongs[api]
+# - gunicorn>=23.0.0; platform_system != "Windows"
+# - uvicorn>=0.30.1
 # - ilt-panphon<0.22,>=0.21.4
 #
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -63,69 +68,87 @@ idna==3.11
...
+    # via
+    #   hatch.envs.prod
+    #   readalongs
 pyyaml==6.0.3
     # via
     #   g2p
-    #   panphon
-regex==2025.10.23
+    #   ilt-panphon
+readalongs==1.2.2
```
dhdaines: Not sure about this, it looks like readalongs depends on itself?  How did you generate the requirements.txt file?

joanise: I followed the steps documented in readme-heroku.md, but I agree, that's weird and should not be there.

joanise:
@dhdaines Alright, take 2. I figured out I had declare the prod features incorrectly in pyproject.toml. I had `heroku` depending on `readalongs[api]` instead of having prod declare `features = [ "heroku", "api" ]`.

I regenerated the reqs file again and now it looks like it should. Thanks for flagging the issue!

PR #300, opened 2026-04-21
Apply some picky coding recommendations

joanise:
### PR Goal? 

Actually, the real goal is to push an update to main so that heroku integration, which was just cycled, gets exercised.

The changes applied are all things that are no longer needed with recent versions of python:
 - io.open instead of open (was useful for Python 2.6/2.7)
 - redundant file mode when we're using the default mode
 - `"{}".format()` instead of f-strings
 - `__future__` imports where that future is now
 - `-*- coding: utf-8 -*-`

Also remove some dead code from `utils.py` flagged by codecov. That code was long-since unused.

### Fixes? 

picky details fixed by `uvx ruff check --select UP --fix` and a few manual fixups after

### Feedback sought? 

rubber stamping

### Confidence? 

high


roedoejet (approved):
the future is now! rubber stamped

PR #301, opened 2026-04-21
use pytest in CI, and exclude tests from coverage again

joanise:
### PR Goal? 

In previous refactorings, we accidentally started counting coverage of `tests/*` again. Reconfigure that out.

And let's use pytest in CI too, it's now required and the default, so we can start embracing it.

### Confidence? 

high


joanise:
@dhdaines Adding you to reviewers here because you might be happy to see the move fully to pytest, in CI, in `run_tests.py`, everywhere.

I have not started using fixtures for Studio, but I have one fixture in g2p (to launch g2p_studio). A number of tests in Studio could be simplified with fixtures, but that's going to stay lower priority at the moment.

run_tests.py
```diff
@@ -4,154 +4,121 @@
...
-            yield from match[1].split(", ")
-
-
-def describe_suite(suite: TestSuite):
-    full_suite = LOADER.discover(os.path.dirname(__file__))
-    full_list = list(list_tests(full_suite))
-    requested_list = list(list_tests(suite))
+
+SUITES = {
+    "all": [],  # relies on discovery to collect all tests
```
roedoejet, line 28: ya, this is nice

roedoejet (approved):
looks like a good improvement to me. thanks @joanise

PR #307, opened 2026-06-05
build(deps): drop support for Python 3.8

joanise:
### PR Goal? 

Python 3.8 is long EOL, and it prevents us from using `uv lock` on Windows. There is no value in keeping it anymore, so let's drop it.

### Feedback sought? 

confirmation we don't need Python 3.8 support anymore

### Priority? 

normal

### Tests added? 

all well covered

### Confidence? 

high

### Version change? 

no

### Related PR

https://github.com/NRC-ILT/g2p/pull/492


roedoejet (approved):
sounds good thanks @joanise

PR #311, opened 2026-09-08
convert all tests to pytest

joanise:
### PR Goal? 

Wrapped up work mostly done in July: convert all tests to pytest, no more use of unittest at all.

As you can see in the commit logs, I used various regular expressions and some manual conversions in July, which was more efficient than all manual, but still time consuming. Now I decided to experiment with Codex on a simple task, and finished it all very quickly with the help of GPT 5.6 Luna, with a full line-by-line review of all proposed changes, rejecting and/or fixing what I didn't agree with.

### Fixes? 

Out with unittest!

### Feedback sought? 

normal review

### Priority? 

low

### Tests added? 

nothing but!

### How to test? 

`uv run pytest` 

### Confidence? 

high

### Version change? 

no


tests/test_g2p_cli.py
```diff
@@ -474,21 +462,20 @@ def test_invalid_langs_in_xml(self):
...
+        c_xml, valid, _ = convert_xml(xml, verbose_warnings=True)
+        assert not valid
+        logger_output = caplog.text
+        assert "No lang" in logger_output
+        assert "foo" in logger_output
+        assert 'no path from "crx-syl"' in logger_output
 
     def test_non_convertible_words(self):
         xml = parse_xml("<s><w>43:23</w><w>65:67</w><w>43:23</w></s>")
-        with self.assertLogs(LOGGER, level="WARNING"):
```
roedoejet, line 487: any reason this is removed?

tests/test_anchors.py
```diff
@@ -28,15 +28,15 @@ def test_anchors_inner_only(self):
...
-        self.assertLessEqual(words[0]["end"], 1.62)
-        self.assertGreaterEqual(words[1]["start"], 1.62)
-        self.assertLessEqual(words[8]["end"], 3.81)
-        self.assertGreaterEqual(words[9]["start"], 3.82)
-        self.assertLessEqual(words[21]["end"], 6.74)
-        self.assertGreaterEqual(words[22]["start"], 6.74)
+        assert words[0]["end"] <= 1.62
+        assert words[1]["start"] >= 1.62
+        assert words[8]["end"] <= 3.81
+        assert words[9]["start"] >= 3.81
```
roedoejet: why'd this get changed from 3.82 to 3.81?

roedoejet (approved):
Approving, but I think there might be a couple very small issues. please look at the comments and fix or ignore as you see fit. Thanks Eric, this looks great!

tests/test_g2p_cli.py
joanise: This is one of the many cases where I was using assertLogs to hide the massive output we used to get to terminal with unittest. I've never cared whether this logged, and still don't, the relevant assertions are on the next two lines.

joanise: With pytest, it captures everything for us, so all the trouble I went to to reduce output is no longer needed. And by removing this nonsense, should you want to see that output, you are able to get it with `pytest -s`, instead of having to change the test code to stop hiding it.

tests/test_anchors.py
joanise: Good catch, I must have thought it was a typo when I converted, but in fact the input data file has two anchors, one at 3.81 and one at 3.82, and the original assertions were correct. I'll fix this one.

